PR #3, opened 2016-04-03
Merge in documentation changes and namespace cleanup

nholthaus:


nholthaus:
This needs to be worked on with vs2013 before pulling.

PR #30, opened 2016-11-03
Fix typo _MCS_VER -> _MSC_VER and only add pragma warning when it's actually defined

pvaibhav:
Trying to compile on GCC-ARM 5.4.1, the `#pragma warning` generates a warning that the pragma is being ignored, presumably because this only applies to MSVC. I've corrected the existing typo and added another check for the macro actually being defined. This gets rid of the warnings related to the pragma.

Note though I'm still unable to build on GCC-ARM due to unrelated errors/warnings which I'll be fixing later.

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -50,7 +50,7 @@
 #ifndef units_h__
 #define units_h__
 
-#if _MCS_VER < 1800
+#if defined(_MSC_VER) && _MSC_VER < 1800
 #	pragma warning(push)
 #	pragma warning(disable : 4520)
 #endif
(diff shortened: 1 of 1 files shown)


nholthaus:
Looks good. Thanks for your contribution! ARM isn't a technology that I typically use, so I'll be excited to pull in your port.

If you plan to make more contributions you may want to work off the v2.2.x branch. The development is basically finalized and I'm planning to release it as v2.2.0 within the next couple of weeks.

pvaibhav:
Thanks for merging. I'll pull in v2.2.x now.

PR #34, opened 2016-11-06
Fix -Wpedantic warnings

Oxyd:
This allows usage of this library in projects that compile with -std=c++14 -Wall -Wextra -pedantic. Tested on GCC 6.2.1.


nholthaus:
Looks good. I think I'll also enable the higher levels of warning in future testing.

PR #36, opened 2016-11-06
README.md: fix conversion factor in comment

dharmatech:


nholthaus:
@dharmatech Thanks for all your help! Your contributions in terms of issues, pull requests, and getting the word out about the library are invaluable. Having such engaged and enthusiastic users really makes the effort worthwhile!

PR #42, opened 2017-02-06
Change MACRO definitions to support adding units with/without IO/literals. Unify VS versions.

JaapAap:
The MACRO definitions that generate the unit types were updates to support adding units with or without IO, and with or without literals. This also allows to support different compilers (or compiler versions) by enabling/disabling these macros. 

Literals are implemented using a `make_unit<T>(value)` function to construct a unit from a plain double. This seems to be a useful function on its own, as it ensures the user has to explicitly provide the type right next to the plain value, which is not necessarily the case when using an (explicit) constructor (for example in member initialization).

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -302,13 +302,10 @@
 /**
  * @def		UNIT_ADD_WITH_METRIC_PREFIXES(nameSingular, namePlural, abbreviation, definition)
  * @brief	Macro for generating the boiler-plate code needed for a new unit, including its metric
- *			prefixes from fempto to peta.
- * @details	This macro should be used within an appropriate namespace for the unit
- *			category. The macro generates singular, plural, and abbreviated forms
- *			of the unit definition (e.g. `meter`, `meters`, and `m`), as well as the
- *			appropriately named unit container (e.g. `meter_t`). A literal suffix is created
- *			using the abbreviation (e.g. `10.0_m`). It also defines a class-specific
- *			cout function which prints both the value and abbreviation of the unit when invoked.
+ *			prefixes from femto to peta.
+ * @details	See UNIT_ADD. In addition to generating the unit definition and containers '(e.g. `meters` and 'meter_t', 
+ *       it also creates corresponsing units with metric suffixesm such as `millimeters`, and `millimeter_t`), as well as the
+ *			literal suffixes (e.g. `10.0_mm`).
  * @param	namespaceName namespace in which the new units will be encapsulated. All literal values
  *			are placed in the `units::literals` namespace.
  * @param	nameSingular singular version of the unit name, e.g. 'meter'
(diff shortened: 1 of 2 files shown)


nholthaus:
From a cursory visual inspection, this looks like it's refactored exactly the way I intended. I think the gcc/clang errors aren't a big deal and I can clean them up along with some small style things. I have a few questions though to speed up the process of [groking](https://en.wikipedia.org/wiki/Grok) the changes and pulling them faster:

- How were you able to remove the base unit definition from the `is_<category>_unit` macro definitions?
- Why'd you comment out the `constexpr` test?
- I don't think I have any problem or objection to the `make_unit` function, but I'm also not sure I understand the use case. Could you provide an example?

Most importantly, thanks for contributing!!

nholthaus:
@nholthaus reminder to self: create CMake option to disable IOStream, edit readme.

include/units.h
```diff
@@ -71,165 +76,192 @@
...
+* @param	namespaceName namespace in which the new units will be encapsulated. All literal values
+*			are placed in the `units::literals` namespace.
+* @param	nameSingular singular version of the base unit name, e.g. 'watt'
+* @param	abbreviation - abbreviated decibel unit name, e.g. 'dBW'
+*/
+#define UNIT_ADD_DECIBEL(namespaceName, nameSingular, abbreviation)\
 	namespace namespaceName\
 	{\
-		/** @name Unit Containers */ /** @{ */ typedef unit_t<nameSingular, UNIT_LIB_DEFAULT_TYPE, units::decibel_scale> abbreviation ## _t; /** @} */\
+		/** @name Unit Containers */ /** @{ */ using abbreviation ## _t = unit_t<nameSingular, UNIT_LIB_DEFAULT_TYPE, units::decibel_scale>; /** @} */\
```
nholthaus: The reason I was using `typedef` instead of aliases is that (interestingly) some compiler errors will actually report descriptive typenames instead of the template definition, but only when it's a `typedef` and not an alias.

include/units.h
```diff
@@ -1946,14 +1989,39 @@ namespace units
 		template<class U, typename Ty, template<typename> class Nlt>
 		friend class unit_t;
 
+	  /**
+	  * @ingroup	???
```
nholthaus: ingroup should be `UnitContainers`

include/units.h
```diff
@@ -1946,14 +1989,39 @@ namespace units
...
 
+	template<class UnitType, typename T>
+	UnitType
+	make_unit(const T value)
+	{
+		static_assert(traits::is_unit_t<UnitType>::value, "Template parameter `UnitType` must be a unit type (_t).");
+		UnitType Unit;
+		Unit.m_value = static_cast<typename UnitType::underlying_type>(value);
+		return Unit;
+	}
```
nholthaus: check that `T` is an arithmetic type. `return UnitType(value);`

unitTests/main.cpp
```diff
@@ -1169,6 +1170,8 @@ TEST_F(UnitContainer, cout)
 	std::cout << test6;
 	std::string output6 = testing::internal::GetCapturedStdout();
 	EXPECT_STREQ("120 dBm", output6.c_str());
+
+	#endif
```
nholthaus: move defines to outside of test scope

JaapAap:
Looks good, thanks! I guess that make_unit does not need to be a ‘friend’ if it is allowed to use the explicit constructor (recall that it was originally conceived as an alternative to that constructor)

-          Pieter

nholthaus:
Thanks for the clarification, I'll make that change as well.

PR #59, opened 2017-02-12
Reduce metaprogramming boilerplate

Morwenn:
I tried to reduce a bit the boilerplate due to template stuff, which tends to make super-long lines that sometimes hinder the readability of the code. Here are the main changes:
* Replace `typename std::type_trait<T>::type` by C++14 `std::type_trait_t<T>` when possible
* Replace `typename std::conditional<Condition, std::true_type, std::false_type>::type` by `std::integral_constant<bool, Condition>`

I was also about to remove `inline` where it is implied by `constexpr`, but realized that `constexpr` was sometime an empty `#define`, so keeping `inline` makes sense.

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -1076,8 +1076,8 @@ namespace units
 
 			template <intmax_t Lower, intmax_t Upper, typename Condition1 = void, typename Condition2 = void>
 			struct DoubleSidedSearch_ : DoubleSidedSearch_<Lower, Upper,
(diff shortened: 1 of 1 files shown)


nholthaus:
Love it. However, would you mind opening another pull request, and merging into the development branch, `v2.3.x` instead?

Morwenn:
Sure, I'm on it :)

PR #60, opened 2017-02-12
Reduce metaprogramming boilerplate

Morwenn:
Same as #59, but rebased on v2.3.x

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -280,14 +280,14 @@
 #define UNIT_ADD_IS_UNIT_CATEGORY_TRAIT(unitCategory)\
 	namespace traits\
 	{\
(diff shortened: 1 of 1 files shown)


nholthaus:
I'll take a look into the internal compiler error. _Some_ of the boilerplate represents patterns that I discovered over time which VS2013 could parse better than others. My bet is it's in the `UNIT_CATEGORY_TRAIT` macro area.

Morwenn:
Ugh, I'm only testing with MinGW at home, and tend to forget that old VS versions have problems with what looks like not-so-hard template metaprogramming. I didn't think such simple changes would trigger an ICE :/

Well, it's still cool that you managed to fix it.

Issue #67, opened 2017-03-03
Compiler error when UNIT_LIB_DEFAULT_TYPE is changed to float


pvaibhav:
When I changed the `UNIT_LIB_DEFAULT_TYPE` to `float`, my compiler (gcc-arm) spits out a very long error basically boiling down to:

```
units/include/units.h:1452:28: error: '(6.022141e+23f * 6.022141e+23f)' is not a constant expression
    return y == 0 ? 1.0 : x * pow(x, y - 1);
                          ~~^~~~~~~~~~~~~~~
```

It took a long time to realize the reason: (6.022141e+23f * 6.022141e+23f) is of the order of 10^46, which doesn't fit in a single-recision float whose max value is around 10^38.

A little more debugging later turns out the derivation of Stefan-Boltzmann constant around line 3954:

```
static constexpr const unit_t<compound_unit<power::watts, inverse<area::square_meters>, inverse<squared<squared<temperature::kelvin>>>>>
	sigma((2 * math::cpow<5>(pi) * math::cpow<4>(R)) / (15 * math::cpow<3>(h) * math::cpow<2>(c) * math::cpow<4>(N_A)));	///< Stefan-Boltzmann constant.
```

contains Avogadro's number raised to 4th power.

I replaced this with pre-calculated constant, and the error disappeared:

```
static constexpr const unit_t<compound_unit<power::watts, inverse<area::square_meters>, inverse<squared<squared<temperature::kelvin>>>>>
	sigma(5.670367e-8);///< Stefan-Boltzmann constant.
```

achary:
Personally I find this bug quite annoying. It is still present and prevents using the library correctly (without patching) with UNIT_LIB_DEFAULT_TYPE float.
I would suggest to re-visit the "wontfix" decision for it and replace Stefan-Boltzmann (sigma) constant with a fixed numerical value, rather then compile-time computing it from others.

nholthaus:
I'd marked this wontfix because the issue is solved generically in the still-unreleased version 3, which I'd hoped to have released about a month ago. In v3 we support templated underlying types and mixing/matching them properly, so the `UNIT_LIB_DEFAULT_TYPE` isn't needed by users.

In general I want to be pedantic about not approximating units or constants, but you've convinced me in this case.

GElliott:
v3 also leaves C++14 users behind.  I would like to see better support for `float` in v2.

johelegp:
I wouldn't mind making v3 work with C++14.

[...]

PR #68, opened 2017-03-03
Fix Issue #67 (compiler error when using float as underlying type)

pvaibhav:
Replace compile-time derived value of Stefan-Boltzmann constant with
pre-calculated value. This prevents compiler error due to the derived
value calculation containing Avogadro's number (order 10^23) raised to
4th power, which doesn't fit in a single precision float (max order
10^38). See Issue #67 for more information.


nholthaus:
I don't want to define derived units as a rule, so I think I'll leave it as an exercise to the user to disable constants/change them on their own if they mess with the underlying type. Thanks for the info though!

pvaibhav:
>I don't want to define derived units as a rule

May I ask why an exception cannot be made for physical *constants*?

>I think I'll leave it as an exercise to the user to disable constants/change them on their own if they mess with the underlying type.

I think such a strict rule is very user-unfriendly, specially without a strong reason justifying that failing to compile is more preferable to this.

Consider that the underlying type is one of the 3 fundamental components that define a unit in this library, and changing the underlying type seems to be supported and documented. However, changing from `double` to `float` instantly breaks compilation – arguably the only other data type most users would want to use.

nholthaus:
@pvaibhav I respect your opinion and you've been one of my more active users, so I'm going to look into how I can more gracefully degrade the constant definitions with changes in underlying type.

PR #72, opened 2017-04-21
add inline to pow() and abs() for different compilation units

stefan-muc:
By using units.h on different compilation units with Visual Studio 2013 I got linker errors like
`2>Second.obj : error LNK2005: "double __cdecl units::detail::pow(double,unsigned __int64)" (?pow@detail@units@@YANN_K@Z) already defined in First.obj`

Although the standard says `(§7.1.5/2): "constexpr functions and constexpr constructors are implicitly inline (7.1.2)." ` ([source](http://stackoverflow.com/questions/14391272/does-constexpr-imply-inline)) it seems that VS2013 doesn't do that.

So I would suggest adding the keyword manually. For me it now compiles like a charm

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -1,34 +1,34 @@
 //--------------------------------------------------------------------------------------------------
-//
+// 
 //	Units: A compile-time c++14 unit conversion library with no dependencies
 //
 //--------------------------------------------------------------------------------------------------
 //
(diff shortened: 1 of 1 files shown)


nholthaus:
The branch is now in good shape if you want to make another attempt.

PR #73, opened 2017-04-21
add inline to pow() and abs() for different compilation units #2

stefan-muc:
New try for [the same issue](https://github.com/nholthaus/units/pull/72)

So here am I :)

Sorry for the inconvenience, my editor automatically removes all whitespace from end of the line - seems as you had much of it.


nholthaus:
yeah the whitespace is a problem which I'll clean up in its own issue.

PR #74, opened 2017-05-10
Fix error from GCC 4.8.5 on RHEL/Centos 7

vgplayer:
On GCC 4.8.5 from RHEL 7 I get:
error: missing space between '""' and suffix identifer

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -171,11 +171,11 @@
 	#define UNIT_ADD_LITERALS(namespaceName, nameSingular, abbreviation)\
 	namespace literals\
 	{\
-		inline constexpr namespaceName::nameSingular ## _t operator""_ ## abbreviation(long double d)\
+		inline constexpr namespaceName::nameSingular ## _t operator"" _ ## abbreviation(long double d)\
 		{\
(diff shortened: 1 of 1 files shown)


nholthaus:
So that's actually not a bug, it's a C++14 feature (that gcc 4.8.5 doesn't have implemented). From [cppreference.com](http://en.cppreference.com/w/cpp/language/user_literal):

> **operator** _user-defined-string-literal (since c++14)_
>  the character sequence "" followed, without a space, by the character sequence that becomes the ud-suffix. This special syntax makes it possible to use language keywords and reserved identifiers as ud-suffixes, and is used by the declaration of operator ""if from the header <complex>. Note that using this form does not change the rules that user-defined literal operators must begin with an underscore: declarations such as operator ""if may only appear as part of a standard library header. However, it allows the use of an underscore followed by a capital letter (which is otherwise a reserved identifier) 

The minimum officially supported gcc compiler is 4.9, because the library uses a lot of capital-letter literals.

vgplayer:
Understood.  Unfortunately stuck with gcc 4.8.5 until RHEL adds it to their main repo.

PR #84, opened 2017-09-16
added support for cmake install

dinocore1:
To have cmake install units library to a local 'install' directory:

mkdir build
cd build
cmake -DCMAKE_INSTALL_PREFIX="install" ..
cmake --build . --target install

The units library can then be used in some other cmake project using the
standard
'find_package' command. Like so:

find_package(units)


nholthaus:
Thank you very much for this! I wouldn't be surprised if it's the favorite new feature of v2.3.

Issue #91, opened 2017-11-06
Make a Units-config.cmake file with exported targets for lib dependency.


tarberd:
Modern cmake expects you to find your dependencies using find_package(). 
For a simple binary executable, adding the subdirectory works just fine, but when bulding a cpp Library with shared objects and its headers depends on units.h, its mandatory the installation of units.h on system. Find_packege support would be awsome so any one can make install it, and find the header imported target as folows: 
`find_package(Units REQUIRED CONFIG)`
`add_library(foo foo.cpp)`
`target_include_directories(foo PUBLIC ${include_dir})`
`target_link_libraries(foo PUBLIC Units::units)`

tarberd:
Never mind ... had an old

PR #92, opened 2017-11-06
Add namespace to exported target

tarberd:
#91 
I had an old branch as subdirectory in my project, sorry for the misleading issue. Added a few lines to the readme better explaining how to use find_package, and added a namespace to the exported target in order to avoid conficts with user targets.


nholthaus:
great PR, thanks!

Issue #93, opened 2017-11-23
Consider changing the include path for the library


tarberd:
Although using the lib as `#include <units.h>` it's confortable, for a library to be installed on a user's system, it is a better practice to put te lib inside a folder so it would be like `#include <units/units.h>` just like `#include <gtest/gtest.h>`

nholthaus:
have you run into a conflict?

tarberd:
Ney... I've not run into conflict, it's just a mather of consistency. If you look inside the header directorie in linux at `/usr/include` you will be able to see how quickly a mess it can become. I'd reather not contribute to the mess it already is, and keep all package headers inside a folder with the project name. I know it breaks compatibility, hope you consider the change!
Btw, I packaged my fork with this changes already to archlinux and ubuntu.
arch: [tarberd-units-git](https://aur.archlinux.org/packages/tarberd-units-git/)
ubuntu: [units](https://packagecloud.io/tarberd/ophidian)

MikeGitb:
Just one comment regarding collisions (I had one with a non-public library): "units/units.h" is just as generic as "units.h". Would be great, if there was some less generic name/abbreviation that could be used - maybe as simple as `nunits`?

tarberd:
Changing the project name seems to much for me. The name units is already stablished in the code. Changing the name would mess with the namespace. I'd reather have `#include <units/units.h>` matching `using units::foo;` then have the namespace difer from the project name.

MikeGitb:
Has the include path been changed? Or does this mean it will stay as it is? Imho that would be the worst solution.

PR #95, opened 2017-11-23
Changed header location to units/units.h

tarberd:
#93 This will keep things with a more professional look.


nholthaus:
closing because our whole header strategy is different in 1.3

Issue #94, opened 2017-11-23
Suport units.version file.


tarberd:
cmake has suport for a version file, I would suggest using it in order to suport things like:
```
find_package(units 2.3.1 REQUIRED)
```
so if end user do not have a compatible version, find_package wil give an error.

tarberd:
I have also changed the path of the installed `unitsConfig.cmake` file to `share/units/cmake`, since [fin_package](https://cmake.org/cmake/help/latest/command/find_package.html) does not search on `<prefix>/cmake`.

PR #96, opened 2017-11-23
Units version file

tarberd:
#94 Now cmake generates a unitsConfigVersion.cmake file to suport `find_package(units 2.3.1 REQUIRED)` with same major version compatibility :)


tarberd:
Is anything wrong with appveyor ?

PR #98, opened 2018-02-01
Add more #if guards for DISABLE_PREDEFINED_UNITS

Erroneous1:
If you prefer I can resubmit with consolidated #if guards instead of one for each function.


nholthaus:
looks good, I just have to fix the CI system

PR #100, opened 2018-03-05
Added units and constants used in vacuum physics.

bennygit:


nholthaus:
Version 2.3.1 will include metric prefixes for `bar` units, but based on some cursory research, it seems that there isn't a single internationally accepted standard for STP (https://en.wikipedia.org/wiki/Standard_conditions_for_temperature_and_pressure), and they're standard definitions, not empirical physical constants anyway.

It seems like you'd be better off defining a class/tuple of STP values enumerated off of the standard organization that defines them.

PR #111, opened 2018-04-26
variadic dimensions

nholthaus:


include/units.h
```diff
@@ -2442,245 +2430,263 @@ namespace units
 	//	LINEAR ARITHMETIC
 	//------------------------------
 
-	template<class UnitTypeLhs, class UnitTypeRhs, std::enable_if_t<!traits::is_same_scale<UnitTypeLhs, UnitTypeRhs>::value, int> = 0>
-	constexpr inline int operator+(const UnitTypeLhs& /* lhs */, const UnitTypeRhs& /* rhs */) noexcept
+	template<class UnitTypeLhs, class UnitTypeRhs>
+	constexpr inline std::enable_if_t<!traits::has_same_scale_v<UnitTypeLhs, UnitTypeRhs>, int>
```
johelegp: These `operator+` aren't mutually exclusive. In particular, it becomes ambiguous when compatible units have different underlying types due to the requirement on this line, although the other overload does the right thing.

PR #121, opened 2018-05-11
Documentation and code fixes and simplifications

### LINKED ISSUE #115: `has_num` and `has_den` encourage bad practice
The problem with `has_num` and `has_den` is that they are hardly useful on their own, yet are part of the public interface of the library. It is like the `Addable` concept which only tests for the presence of a binary `operator+`, when what we really want is a `Number`. `std::string` matches `Addable`, but not with the expected semantics of a `Number`.

We already have what we really want, `is_ratio_v`, so I suggest dropping `has_num` and `has_den` from the public interface for v3.0.0. I'm already working on some patches which include these changes, but I'm opening an issue anyways. The patch is a simplification of `is_ratio` that depends on the library relying only on `std::ratio` and not a `Ratio` implementation anymore which I found digging through the commits.

johelegp:
64548a4 resolves #115.

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -2095,7 +2095,7 @@ namespace units
 	//-----------------------------------------
 
 	template<class D, class E>
-	inline std::ostream& operator<<(std::ostream& os, const dim<D, E>&) noexcept
+	inline std::ostream& operator<<(std::ostream& os, const dim<D, E>&)
 	{
 		if constexpr(E::num != 0)os << ' ' << D::abbreviation;
 		if constexpr(E::num != 0 && E::num != 1) { os << "^" << E::num; }
@@ -2103,13 +2103,13 @@ namespace units
 		return os;
 	}
 
-	inline std::ostream& operator<<(std::ostream& os, const dimension_t<>&) noexcept
+	inline std::ostream& operator<<(std::ostream& os, const dimension_t<>&)
 	{
 		return os;
 	}
 
 	template<class Dim, class... Dims>
-	inline std::ostream& operator<<(std::ostream& os, const dimension_t<Dim, Dims...>&) noexcept
+	inline std::ostream& operator<<(std::ostream& os, const dimension_t<Dim, Dims...>&)
 	{
 		os << Dim{};
 		os << dimension_t<Dims...>{};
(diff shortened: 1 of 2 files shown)


include/units.h
```diff
@@ -956,7 +904,7 @@ namespace units
 		static_assert(traits::is_ratio_v<PiExponent>, "Template parameter `PiExponent` must be a `std::ratio` representing the exponents of Pi the unit has.");
 		static_assert(traits::is_ratio_v<Translation>, "Template parameter `Translation` must be a `std::ratio` representing an additive translation required by the unit conversion.");
 
-		using dimension_type = units::dimension_t<Exponents...>;
+		using dimension_type = dimension_t<Exponents...>;
```
johelegp: Just above, in the template parameter list, `dimention_t` isn't prefixed with `units::`. Should I have done the change the other way around?

include/units.h
```diff
@@ -993,10 +941,9 @@ namespace units
 	namespace traits
 	{
 		/**
-		* @brief		Trait which returns the `dimension` type that a unit is originally derived from.
-		* @details		Since units can be derived from other `unit` types in addition to `dimension` types,
-		*				the `dimension_type` typedef will not always be a `dimension` (or unit dimension).
-		*				Since compatible
```
johelegp: Did you want to say something on this line? It was incomplete so I got rid of it.

include/units.h
```diff
@@ -1934,7 +1849,7 @@ namespace units
 		}
 
 		/**
-		 * @brief		copy constructor (converting)
```
johelegp: The special member functions are never templates.

johelegp:
Forgive all the commits. I broke up a commit into all its constituents. As I suspected, using `is_arithmetic_v` for SFINAE in the member functions of `unit` was the cause of some Visual Studio errors. Should I regroup the commits?

include/units.h
nholthaus: nope, it was inconsistent and verbose for no reason.

nholthaus:
I squash merge so don't worry about the commits.

This is beautiful, and documentation is usually a long-pole for `units` releases because it's not fun. Thank you!

I left a few small, nit-picky comments. Let me know what you think. Once we figure those out I'll be excited to pull this.

johelegp:
It's a pleasure.

I can't find your comments. You should do the reviews at the [files changed]https://github.com/nholthaus/units/pull/121/files).

johelegp:
Done. I accidentally deleted my review comments. Did you get the answer to https://github.com/nholthaus/units/pull/121#discussion-diff-187743773R2142?

PR #123, opened 2018-05-11
Apply std::chrono::duration_cast optimization to convert

johelegp:
See https://wg21.link/time.duration.cast#2.


nholthaus:
YES. It has so many lesser optimizations, can't believe it missed that. Probably combinatorics from the pre-`if constexpr` days.

Issue #124, opened 2018-05-11
Show off the power of the new variadic dimensions system


nholthaus:
We have user-definable variadic dimensions now, but we need to develop some case studies we can document to guide users on how to take advantage of that. We probably also want some macros to simplify including new dimensions and making sure traits and things are auto-generated when a new dimension is defined.

If anyone comes up with something cool, post it here or in a PR and I'll include it in the docs w/ attribution.

johelegp:
I'm creating a game and its engine. Right now, it only has an "[initial commit](https://github.com/johelegp/Made_in_Abyss)" with an empty `::main` function. While working on the next commit, "open the game window," I created a type [`Pixels`](https://github.com/johelegp/Made_in_Abyss/blob/69aaa78db360ec845c2e533d2be0983e1451081b/include/mia/spacial.hpp#L22) to represent a quantity of pixels. With it as the base of a [strongly typed interface](https://arne-mertz.de/2016/11/stronger-types/), opening the game window would look like
```C++
Window main{Size{Width{800_px}, Heigh{600_px}}};
```
Looking ahead to representing the positions of "entities" in the game's world, I decided on this library's `length::meter_t`. I found it a shame that my units couldn't participate in the dimensional analysis operations of this library. That'd be at least immediately useful for mapping the positions of things in the game's world to the game's window (a `length::meter_t` to `Pixels` mapping) in the expressive way that this library allows.

Then came the work on v3.0.0, which allows user-defined dimensions. I came up with something like this to replace `Pixels`:
```C++
namespace dimension {

    struct Display_length_tag {
        static constexpr gsl::czstring<> name{"display length"};
        static constexpr gsl::czstring<> abbreviation{"px"};
    };

    using Display_length = units::make_dimension<Display_length_tag>;

} // namespace dimension

namespace display_length {

    // Represents a quantity of pixels.
    template <class Arithmetic>
    using Pixels = units::unit<
        units::unit_conversion<std::ratio<1>, dimension::Display_length>,
        Arithmetic>;

} // namespace display_length

using display_length::Pixels;

inline namespace literals {
    inline namespace pixels_literals { //

        constexpr Pixels<double> operator""_px(long double px) noexcept
        {
            return Pixels<double>{gsl::narrow_cast<double>(px)};
        }

        constexpr Pixels<int> operator""_px(unsigned long long px) noexcept
        {
            return Pixels<int>{gsl::narrow_cast<int>(px)};
        }

    } // namespace pixels_literals
[...]

Issue #40, opened 2016-12-29
Typedef notation for underlying number types


EvanBalster:
Hello!

As an option for the library, it would be quite useful to generate typedefs for multiple number formats in one project.  Currently the library only defines `_t`-suffixed typedefs with an underlying type defined by UNIT_LIB_DEFAULT_TYPE.  I propose supporting common or possibly user-defined number types with other type suffixes.  Macro-generated names might look like this:

```
using meter_i = unit_t<meter, int>;
using meter_f = unit_t<meter, float>;
using meter_d = unit_t<meter, double>;
```

As an example use-case, I might represent (absolute) points in a simulated world using fixed-point values at the millimeter scale, and vectors using floating-point values at the meter scale.  I may have certain numerically-unstable math here and there which I wish to promote to double-precision, in a project otherwise using single precision.

I suspect this should be optional, and disabled by default, to avoid code bloat.

nholthaus:
I have some mixed feelings about this, although I agree that you are right that there are use-cases for non `double` underlying types.

## floating point

For floating point, you really only have a choice between `float` and `double`. I did some experimentation with `long double` in a geodesy library I built on top of `units`, and found that the rounding errors were significantly more of a problem then when using `double`, while the additional precision didn't buy anything (presumably because `double` can use the extra register space to store data that will later be truncated, while `long double` can't). My opinion is that there's not a use case for it, at least on x64.

`float` can definitely be useful for gaming or CUDA/GPU processing, but I think in most cases you'd want to set `UNIT_LIB_DEFAULT_TYPE` because either you don't care about the precision or it's dictated by your hardware. While valid, I'm not sure your use-case would work as well as you think, because mixing units with different underlying types (especially int/double) isn't always possible (at least as currently implemented). It's an unfortunate by-product of using the strong typing to enforce dimensionality but having the underlying type be a part of the type definition as well (which we don't care so much about enforcing). Some half-hearted attempts I made to get around the typing problem weren't successful, although my instinct is that it could be done. _I'd definitely take a PR for it if someone was sufficiently motivated._

## integral

The only time I personally have seen the need for integer types is when interfacing with hardware registers when I wanted to be absolutely sure that the bits being written were exactly the way I wanted them (surprisingly, units is actually pretty friendly for embedded applications). In those cases I just made a quick alias `using meter_i = unit_t<meter, int>;` and left the `UNIT_LIB_DEFAULT_TYPE` as `double` (to avoid integer division).

[...]

PR #126, opened 2018-05-12
Support integral underlying types

### LINKED ISSUE #125: Effort to support integral underlying types
There is interest in having the library support integral underlying types. I believe that the [`duration`](https://wg21.link/time.duration) class template of [`std::chrono`](https://wg21.link/time.syn) sets a good example in doing so. Its interface is correct, efficient, easy to use right, and hard to misuse. I propose that we adopt the ways of `std::chrono::duration` to make this library more generally useful.

To properly support integral underlying types, the following points need to be implemented. First comes the starting point of the `unit` objects lifetime, the constructors and its dependent utilities. Then `std::common_type`, the base of safe conversions between units of the same dimensions.

- The underlying type parameter of `unit` constructors [[1](https://github.com/nholthaus/units/blob/9b66a12e42e9dffde3be8c0d3eab2bbe5bd0298a/include/units.h#L1908), [2](https://github.com/nholthaus/units/blob/9b66a12e42e9dffde3be8c0d3eab2bbe5bd0298a/include/units.h#L1919)] should be a template parameter of the constructors that only allows [non-lossy conversions](https://wg21.link/time.duration.cons#1) [ _Note:_ `duration` uses `treat_­as_­floating_­point_­v` because it supports arithmetic-like types, for which the standard traits can't be specialized. To keep this effort simple, I think we should continue using the standard traits. -- _end note_ ].
- The [converting constructor](https://github.com/nholthaus/units/blob/9b66a12e42e9dffde3be8c0d3eab2bbe5bd0298a/include/units.h#L1942) should [prevent truncation and lossy conversions](https://wg21.link/time.duration.cons#4).
- The [`convert`](https://github.com/nholthaus/units/blob/9b66a12e42e9dffde3be8c0d3eab2bbe5bd0298a/include/units.h#L1561) function, used by the converting constructor:
  * Needs to [carry intermediate computations in the widest representation](https://wg21.link/time.duration.cast#3).
  * Should encourage type safety by having a type safe interface, or an alternative provided that has a type safe interface. A type safe interface would have `unit`s in the parameter and return types. If an alternative is provided, `convert` should work with `unit`s instead of `unit_conversion`s because `unit`s know the specific underlying types being handled, allowing `convert` to have a correctly specified parameter and return types.
- Partially specialize [`std::common_type`](https://wg21.link/time.traits.specializations). This is pretty straight forward for linear units that don't need a translation, just like `duration`. I'm not sure how to go about it for other units.
- Provide support for the modulus operators.

This list is a work in progress. I'm just letting this out so that discussion can start.

johelegp:
Resolves #125.

- [x] Prevent lossy initialization of units from arithmetic types.
- [x] Prevent lossy conversions and truncation of units in its converting constructor.
- [x] Provide an alternative to `convert` that carries intermediate computations in the widest representation and has a type safe interface.
- [x] Specialize `std::common_type` for units of the same dimension. The common type should be the least precise unit that both units can be exactly converted to.
- [x] Implement the modulus operators with the new framework.
- [x] Use the new framework in the equality operators.
- [x] Use the new framework in the relational operators.
- [x] Use the new framework in the arithmetic linear operators.
- [x] Use the new framework in the arithmetic decibel operators.
- [x] Use the new framework in the cmath functions.

Along the way, some of the supporting traits and other utilities might need to be changed to support mixed underlying types.

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -67,6 +67,7 @@
 #include <cmath>
 #include <limits>
 #include <numeric>
+#include <utility>
 
 #if !defined(UNIT_LIB_DISABLE_IOSTREAM)
 #include <iostream>
@@ -1466,27 +1467,65 @@ namespace units
 	//	CONSTEXPR MATH FUNCTIONS
 	//------------------------------
 
+	/** @cond */	// DOXYGEN IGNORE
+	namespace traits
+	{
+		// forward declaration
+		template<class T>
+		struct is_unit;
+	}
+
+	namespace detail
+	{
+		/**
+		 * @brief		Helper trait to promote integers or integral units to `double` (units).
+		 * @details		Simulates the promotion undergone by integers when calling the standard cmath functions
+		 *				overloaded on `float`, `double` and `long double`. Works for both arithmetic types and
+		 *				unit types.
+		 */
+		template<typename T, bool IsUnit = false>
+		struct floating_point_promotion : std::conditional<std::is_floating_point_v<T>, T, double> {};
+
+		template<typename T>
+		using floating_point_promotion_t = typename floating_point_promotion<T, traits::is_unit<T>::value>::type;
+
+		template<template<class, typename, template<typename> class> class Unit, class UnitConversion, typename T, template<typename> class NonLinearScale>
+		struct floating_point_promotion<Unit<UnitConversion, T, NonLinearScale>, true>
+		{
+			using type = Unit<UnitConversion, floating_point_promotion_t<T>, NonLinearScale>;
+		};
+	}
+
 	namespace Detail
 	{
-		double constexpr sqrtNewtonRaphson(UNIT_LIB_DEFAULT_TYPE x, UNIT_LIB_DEFAULT_TYPE curr, UNIT_LIB_DEFAULT_TYPE prev)
+		template<typename T, std::enable_if_t<std::is_floating_point_v<T>, int> = 0>
+		constexpr T sqrtNewtonRaphson(T x, T curr, T prev)
 		{
 			return curr == prev ? curr : sqrtNewtonRaphson(x, 0.5 * (curr + x / curr), curr);
 		}
 	}
+	/** @endcond */	// END DOXYGEN IGNORE
 
-	inline constexpr UNIT_LIB_DEFAULT_TYPE sqrt(UNIT_LIB_DEFAULT_TYPE x)
+	template<typename T, std::enable_if_t<std::is_arithmetic_v<T>, int> = 0>
+	constexpr detail::floating_point_promotion_t<T> sqrt(T x_)
 	{
-		return x >= 0 && x < std::numeric_limits<UNIT_LIB_DEFAULT_TYPE>::infinity()
-			? Detail::sqrtNewtonRaphson(x, x, 0)
-			: std::numeric_limits<UNIT_LIB_DEFAULT_TYPE>::quiet_NaN();
+		using FloatingPoint = detail::floating_point_promotion_t<T>;
+
+		const FloatingPoint x(x_);
+
+		return x >= 0 && x < std::numeric_limits<FloatingPoint>::infinity()
+			? Detail::sqrtNewtonRaphson(x, x, FloatingPoint(0))
+			: std::numeric_limits<FloatingPoint>::quiet_NaN();
 	}
 
-	inline constexpr UNIT_LIB_DEFAULT_TYPE pow(UNIT_LIB_DEFAULT_TYPE x, unsigned long long y)
+	template<typename T, std::enable_if_t<std::is_arithmetic_v<T>, int> = 0>
+	constexpr detail::floating_point_promotion_t<T> pow(T x, unsigned long long y)
 	{
 		return y == 0 ? 1.0 : x * pow(x, y - 1);
 	}
 
-	inline constexpr UNIT_LIB_DEFAULT_TYPE abs(UNIT_LIB_DEFAULT_TYPE x)
+	template<typename T, std::enable_if_t<std::is_arithmetic_v<T>, int> = 0>
+	constexpr T abs(T x)
 	{
 		return x < 0 ? -x : x;
 	}
@@ -1597,13 +1636,6 @@ namespace units
 	}
 
 	/** @cond */	// DOXYGEN IGNORE
-	namespace traits
-	{
-		// forward declaration
-		template<class T>
-		struct is_unit;
-	}
-
 	namespace detail
 	{
 		/**
@@ -2952,10 +2984,10 @@ namespace units
 		* @param[in]	value `unit` derived type to raise to the given <i>power</i>
 		* @returns		new unit, raised to the given exponent
 		*/
-	template<int power, class UnitType, class = typename std::enable_if<traits::has_linear_scale_v<UnitType>, int>>
-	inline constexpr auto pow(const UnitType& value) noexcept -> unit<typename units::detail::power_of_unit<power, typename units::traits::unit_traits<UnitType>::unit_conversion>::type, typename units::traits::unit_traits<UnitType>::underlying_type, linear_scale>
+	template<int power, class UnitType, std::enable_if_t<traits::has_linear_scale_v<UnitType>, int> = 0>
+	constexpr auto pow(const UnitType& value) noexcept -> unit<typename units::detail::power_of_unit<power, typename units::traits::unit_traits<UnitType>::unit_conversion>::type, detail::floating_point_promotion_t<typename units::traits::unit_traits<UnitType>::underlying_type>, linear_scale>
 	{
-		return unit<typename units::detail::power_of_unit<power, typename units::traits::unit_traits<UnitType>::unit_conversion>::type, typename units::traits::unit_traits<UnitType>::underlying_type, linear_scale>
+		return unit<typename units::detail::power_of_unit<power, typename units::traits::unit_traits<UnitType>::unit_conversion>::type, detail::floating_point_promotion_t<typename units::traits::unit_traits<UnitType>::underlying_type>, linear_scale>
 			(pow(value(), power));
 	}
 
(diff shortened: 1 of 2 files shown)


nholthaus:
looking awesome so far.

johelegp:
75eb9ef introduces `unit_conversion_t` as an alias for an `unit_conversion` whose dimension is always a `dimension_t`. This is done to prevent anomalies like `unit_conversion<std::ratio<1>, unit_conversion<std::ratio<1>, dimension::length>>`. This addition flattens the permitted `unit_conversion`s. This allows implementing a specialization of `std::common_type` that Just Works. It also reduces the length of the error messages thanks to the flattening. And maybe other things thanks to the increased consistency.

You'll see the addition of a few macros. The flattening allowed me to discover a few units that before were equivalent, e.g. `meter_t` and `unit<unit_conversion<std::ratio<1>, length::meter>, double>`, but were different types simply because their `UnitType` parameter were different types. The fact that they're now the same type means that the definition of units introduce redefinitions, so the new macros avoid those.

johelegp:
The `unit_conversion_t` approach (c449ade ) didn't quite work. The library defines units like `metric_ton`, which is an alias for `unit_conversion<std::ratio<1000>, kilograms>`. That's equivalent to `megagrams`, but not the same. This allows the library to have a strong type for `megagrams`, which this change would break. Some things I noticed is that `unit_conversion<std::ratio<1>, metric_ton>` counts as an ephemeral unit, and falls back to streaming this unit in terms of `kg` instead of the `t` abbreviation. And that ephemeral units don't name `name` or `abbreviation`.

I couldn't come up with a change that provide the benefits of both. And it wasn't even necessary to do so. I thought it'd be nice to use `std::is_same` in the tests for `std::common_type`, but I'll just use the traits of the library instead.

GElliott:
I've been hacking up changes to v2.3 to make it easier to mix units with underlying types (my use-case is outlined at the end of [this issue](https://github.com/nholthaus/units/issues/40)), and I see some overlap with this work.  For example, this would let you add a float-based meter type to a double-based meter type.  Should we coordinate?

johelegp:
Sure! We could do it here or [in my fork](https://github.com/johelegp/units/tree/duration_like_unit).

johelegp:
I have reread your message (I had done so before, back when I was contemplating someday using this library). The changes I'm going to do will indeed allow you to do 1 and 2, but 3 is not part of this effort. What does the library do that impedes its `unit`s being standard layout?

Regarding your remark about `duration_cast`, in your listed example you can indeed do the assignment from and to `float` and `double` `duration`s (and `unit`s, with the second commit of this PR). However, I agree that `std::chrono` might have gone overboard. We have the `convert` to do these explicit casts, but we could also have explicit constructors for such lossy conversions.

johelegp:
I had been struggling with formulating a specialization of `std::common_type` that worked for units with pi exponent and translation, and only just realized that conceptually their greatest common divisor does the job, just like the regular conversion ratio. I really didn't want to continue working on the other tasks before getting this done, otherwise I feared I'd have to limit the effort to the units without those ratios. Now all that's left is including the non-linear scale in the definition.

include/units.h
```diff
@@ -4379,8 +4385,59 @@ namespace units
...
+	 *				any value of these conversions, although floating-point units may have round-off errors.
+	 *				If the units have mixed scales, preference is given to `linear_scale` for their common type.
+	 */
+	template<class UnitConversionLhs, class Tx, class UnitConversionRhs, class Ty, template<typename> class NonLinearScale>
+	struct common_type<units::unit<UnitConversionLhs, Tx, NonLinearScale>, units::unit<UnitConversionRhs, Ty, NonLinearScale>> :
+		std::enable_if<units::traits::is_convertible_unit_conversion_v<UnitConversionLhs, UnitConversionRhs>,
+			units::unit<
+				units::unit_conversion<
+					units::detail::ratio_gcd<typename UnitConversionLhs::conversion_ratio, typename UnitConversionRhs::conversion_ratio>,
+					typename UnitConversionLhs::dimension_type,
```
johelegp: I wonder if using `dimension_of_t` would be most correct here.

include/units.h
```diff
@@ -1725,6 +1842,28 @@ namespace units
...
+		{
+			static constexpr bool value = std::ratio_divide<typename UnitConversionFrom::conversion_ratio, typename UnitConversionTo::conversion_ratio>::den == 1;
+		};
+
+		/**
+		 * @brief		SFINAE helper to test if a conversion of units is non lossy.
+		 */
+		template <class UnitFrom, class UnitTo>
+		inline constexpr bool IsNonLossyUnitConvertible = std::is_floating_point_v<typename UnitTo::underlying_type> || std::conjunction_v<std::negation<std::is_floating_point<typename UnitFrom::underlying_type>>, detail::no_truncated_unit_conversion<typename UnitFrom::unit_conversion, typename UnitTo::unit_conversion>>;
+
```
nholthaus: Just to nitpick a bit, traits, both types and boolean, should use the `std` underscore convention instead of camel case.

include/units.h
```diff
@@ -1860,9 +1999,9 @@ namespace units
 		 * @details		enable implicit conversions from std::chrono::duration types ONLY for time units
 		 * @param[in]	value value of the unit
 		 */
-		template<class Rep, class Period, class = std::enable_if_t<std::is_arithmetic_v<Rep> && traits::is_ratio_v<Period>>>
+		template<class Rep, class Period, class = std::enable_if_t<std::is_arithmetic_v<Rep>>>
 		inline constexpr unit(const std::chrono::duration<Rep, Period>& value) noexcept : 
-		nls(units::convert<units::unit_conversion<std::ratio<1,1000000000>, dimension::time>, UnitType>(static_cast<T>(std::chrono::duration_cast<std::chrono::nanoseconds>(value).count())))
+		nls(units::convert<unit>(units::unit<units::unit_conversion<Period, dimension::time>, Rep>(value.count()))())
```
nholthaus: so much better!

include/units.h
johelegp: Very well.

include/units.h
```diff
@@ -1912,7 +2051,8 @@ namespace units
 		template<class UnitTypeRhs, typename Ty, template<typename> class NlsRhs>
 		inline constexpr bool operator<(const unit<UnitTypeRhs, Ty, NlsRhs>& rhs) const noexcept
 		{
-			return (nls::m_value < units::convert<UnitTypeRhs, UnitType>(rhs.m_value));
+			using CommonUnit = std::common_type_t<unit, unit<UnitTypeRhs, Ty, NlsRhs>>;
+			return (CommonUnit(*this).m_value < CommonUnit(rhs).m_value);
```
nholthaus: @nholthaus verify that `common_type` conversions have same of improved performance.

include/units.h
```diff
@@ -2041,7 +2191,7 @@ namespace units
 		inline constexpr operator Ty() const noexcept 
 		{ 
 			// this conversion also resolves any PI exponents, by converting from a non-zero PI ratio to a zero-pi ratio.
-			return static_cast<Ty>(units::convert<UnitType, units::unit_conversion<std::ratio<1>, units::dimension::dimensionless>>((*this)()));
+			return units::convert<units::unit<units::unit_conversion<std::ratio<1>, units::dimension::dimensionless>, Ty, NonLinearScale>>(*this)();
```
nholthaus: love this.

include/units.h
```diff
@@ -2151,52 +2301,117 @@ namespace units
...
+			units::unit<
+				units::unit_conversion<
+					units::detail::ratio_gcd<typename UnitConversionLhs::conversion_ratio, typename UnitConversionRhs::conversion_ratio>,
+					units::traits::dimension_of_t<UnitConversionLhs>,
+					units::detail::ratio_gcd<typename UnitConversionLhs::pi_exponent_ratio, typename UnitConversionRhs::pi_exponent_ratio>,
+					units::detail::ratio_gcd<typename UnitConversionLhs::translation_ratio, typename UnitConversionRhs::translation_ratio>>,
+				common_type_t<Tx, Ty>,
+				NonLinearScale>> {};
+}
+
```
nholthaus: looks good!

include/units.h
```diff
@@ -2151,52 +2301,117 @@ namespace units
...
+	constexpr unit<UnitConversion, T, NonLinearScale>& operator+=(unit<UnitConversion, T, NonLinearScale>& lhs, const detail::type_identity_t<unit<UnitConversion, T, NonLinearScale>>& rhs) noexcept
+	{
 		lhs = lhs + rhs;
 		return lhs;
 	}
 
-	template<class UnitConversion, typename T, template<typename> class NonLinearScale, typename RhsType>
-	inline constexpr unit<UnitConversion, T, NonLinearScale>& operator-=(unit<UnitConversion, T, NonLinearScale>& lhs, const RhsType& rhs) noexcept
+	template<class UnitConversion, typename T, template<typename> class NonLinearScale>
+	constexpr unit<UnitConversion, T, NonLinearScale>& operator-=(unit<UnitConversion, T, NonLinearScale>& lhs, const detail::type_identity_t<unit<UnitConversion, T, NonLinearScale>>& rhs) noexcept
```
nholthaus: what's `type_identity_t`?

include/units.h
johelegp: It took me a while to correct this, which was the last use of the non-type safe `convert`.

nholthaus:
still looking beautiful. I almost cried when I saw that you refactored the unit tests away from the VS2013 compliant mess they were before, that was not something I was looking forward to.

include/units.h
johelegp: It's the identity metafunction. It's defined [here](https://github.com/nholthaus/units/pull/126/commits/87194b5854f89438fabf34888066e2a977d81a4c#diff-cc75003c399d878c360873c2c8c783d6R2306) and allows disabling template argument deduction. It's been proposed for standarization. The [paper](https://wg21.link/P0887) explains it well.

johelegp:
That was essential to prove that the type safe `convert` works correctly.

nholthaus:
They were happy tears :)

johelegp:
I think it's done. If anything, I should add some tests for the last five items in the lists ("Use the new framework in"), which only used the existing tests, which don't account for mixed-underlying types units.

Now unused:
- [`traits::has_same_scale`](https://github.com/nholthaus/units/pull/126/files#diff-cc75003c399d878c360873c2c8c783d6L2385)

Further work:
- When the macros for defining units are reworked (#124), they should be underlying type aware. The `operator<<` and `to_string` functions only work on `UNIT_LIB_DEFAULT_TYPE` units, potentially performing lossy conversions. There's also the `operator<<` for `dB` units.
- Consider making integer unit literals return their units with an integer underlying type.
- Investigate having constants of different underlying types. These might be useful for users who can only use `float`, for example. There have been proposals for standarization regarding that.
- Consider deprecating `UNIT_LIB_DEFAULT_TYPE` and substituting its uses with `double`.
Further work:
- When the macros for defining units are reworked (#124), they should be underlying type aware. The `operator<<` and `to_string` functions only work on `UNIT_LIB_DEFAULT_TYPE` units, potentially performing lossy conversions.
- Consider making integer unit literals return their units with an integer underlying type.
- Investigate having constants of different underlying types. These might be useful for users who can only use `float`, for example. There have been proposals for standarization regarding that.
- Consider deprecating `UNIT_LIB_DEFAULT_TYPE` and substituting its uses with `double`.

johelegp:
>still looking beautiful. I almost cried (with joy) when I saw that you refactored the unit tests away from the VS2013 compliant mess they were before, that was not something I was looking forward to.

It can be refactored further. From `convert<millimeter_t>(meter_t(0.001))` to `millimeter_t(meter_t(0.001))`. The later is implemented in terms of `convert` and reflects the guideline of preferring implicit conversions.

PR #132, opened 2018-05-24
Optimize convert with constexpr pi in denominator

johelegp:
I missed this branch because I copy pasted the one above and missed adding `-` to `PiRatio::num`. Having only changed `*` for `/`, the wrong test results led me to believe that it was a problem with the expression being more inaccurate than before.


nholthaus:
Conflicts from the int support branch. I'll pull this once those are worked out.

johelegp:
This PR is subsumed by the accepted PR #126.

PR #142, opened 2018-06-17
Remove warnings

johelegp:


johelegp:
The warnings from MSVC are from testing an intentionally lossy part of the interface.

PR #143, opened 2018-06-20
SI stands for Système International

crisluengo:
SI stands for Système International d'unités, not "Scientific International". https://en.wikipedia.org/wiki/International_System_of_Units


martinmoene:
Perhaps good to add as reference to Readme:

Ambler Thompson and Barry N. Taylor. [Guide for the Use of the International System of Units (SI)](http://physics.nist.gov/cuu/pdf/sp811.pdf). NIST Special Publication 811 2008 Edition.

Issue #133, opened 2018-06-03
User-friendly macros


JohelEGP:
I have started working on improving the macros for defining units to make them usable from outside the library on my branch [johelegp:macros](https://github.com/nholthaus/units/compare/v3.x...johelegp:macros?expand=1). I'd like input on how to proceed on the following points.

The [`UNIT_ADD_NAME`](https://github.com/nholthaus/units/blob/v3.x/include/units.h#L203) assumes that it is being used inside the `::units` namespace, and so do the macros that use it. Since it is a customization point by specialization of the primary templates in `::units`, the correct use from outside the library would consist on using the macro on the global namespace, and wrapping it in `namespace units { ... }`. I don't know why these were introduced, since they're only (even if indirectly) called in `unit`'s `name` and `abbreviature` member functions.

Similarly, [`UNIT_ADD_LITERALS`](https://github.com/nholthaus/units/blob/v3.x/include/units.h#L224) unilaterally places things in the `literals` namespace, which the user might not want. And if they later open than namespace as `inline`, they get an error. The other way around, it's a warning. Probably subject to inclusion order, one day the warning might become an error.

Perhaps a better approach would be to separate the components of [`UNIT_ADD`](https://github.com/nholthaus/units/blob/v3.x/include/units.h#L256) and similar so as to avoid these clashes. Then, their use might look like this:
```C++
namespace my_lib {
  UNIT_ADD(my_units, fortnight, fortnights, ftn, units::unit_conversion<std::ratio<14>, units::time::days>)
  inline namespace literals {
    inline namespace time_literals {
      UNIT_ADD_LITERALS(my_units, fortnight, ftn)
    }
  }
}
UNIT_ADD_NAME(my_lib::my_units, fortnight, ftn)
```
[`UNIT_ADD_DIMENSION_TRAIT`](https://github.com/nholthaus/units/blob/v3.x/include/units.h#L339) has the same problems as `UNIT_ADD_LITERALS`. Also, its [`detail`](https://github.com/nholthaus/units/blob/v3.x/include/units.h#L319) machinery needn't be "exported". It can remain here and have the macro use it.

johelegp:
>Also, its detail machinery needn't be "exported". It can remain here and have the macro use it.
See 4b774b2 to see what I meant.

nholthaus:
[...]

Issue #138, opened 2018-06-15
v3.x: abbreviation & name are broken when using integral underlying types


pattop:
units::unit<meters, int> c;
	EXPECT_STREQ("m", units::abbreviation(c));
	EXPECT_STREQ("meter", units::name(c));
	EXPECT_STREQ("m", c.abbreviation());
	EXPECT_STREQ("meter", c.name());

units.h:109:52: warning: inline function ‘constexpr const char* units::abbreviation(const T&) [with T = units::unit<units::unit_conversion<std::ratio<1>, units::dimension_t<units::dim<units::dimension::length_tag, std::ratio<1, 1> > > >, int>]’ used but never defined

It would also be nice to support querying for name/abbreviation without requiring an object, maybe something like:

	EXPECT_STREQ("ft", units::abbreviation<foot_t>());
	EXPECT_STREQ("foot", units::name<foot_t>());

Another minor thing, in the unit tests the nameAndAbbreviation test case is ifdef'd out by UNIT_LIB_DISABLE_IOSTREAM. That doesn't look quite right.

johelegp:
Ah, that's because they're [defined](https://github.com/nholthaus/units/blob/v3.x/include/units.h#L108-L109) as a template and specialized for the defined units. I guess a more correct approach now that there's support for non-`double` units is to make them overloads.

pattop:
Makes sense.

By the way, nice job on the integral type support. It's going to be really useful.

Off topic, but have you considered whether it would be possible/make sense to use something like https://github.com/johnmcfarlane/cnl as the underlying type?

johelegp:
Thank you.

Strictly, the library only support arithmetic types. However, it's possible to relax this. I don't know if @nholthaus has plans for that.

I see the link is related to the Numbers TS. The answer is in whether it makes sense to use them as an underlying type of `std::chrono::duration`, which only loosely requires it to be arithmetic-like. I'd say your YMMV depending on the assumptions your implementations make of what's arithmetic-like. I guess that'll improve in the future when people start questioning whether the Numeric TS's types can be used with `std::chrono::duration`.

johelegp:
I haven't been successful in finding a way to make `units::name` and `units::abbreviation` work with non-`double` types. [Guidelines](http://www.gotw.ca/publications/mill17.htm) suggest to not use function templates as customization points trough specialization. As we're supporting program-defined units, this customization point should be redesigned accordingly.

nholthaus:
[...]

PR #150, opened 2018-07-07
Strenghten unit aliases

johelegp:
Discussed in #138.

This is a hack to get a feeling of how it would be to make the units actual types and not aliases. This would allow us to use CTAD (which isn't looking hopeful for type aliases in C++20) and better error messages.

My observations:
- The library adds a lot of symbols. I've read about linkers having problems with over 2^16 symbols. This change adds many more.
- This is WIP. e646b14, being the hack it is, adds redundant declarations to get things to work. More symbols.
- Unfortunately, the way `std::common_type` and `std::hash` work require us to add _more symbols_ (specializations) for each strong unit alias.
- The tests pass when I change uses of `std::is_same_v` for `has_equivalent_unit_conversion`.

While working on this, I'll see if I can reduce the amount of symbols the library produces. Making `is_derived_from_unit` (added in this hack) what `is_unit` is and taking into account everywhere of this new definition should help there.

Foreseeable future work includes preserving the strong unit alias whenever possible. As it is, it converts to its base ASAP, losing on the advantage of better error messages.

I just wanted to get this hack out. I'll make it a proper work later.


Morwenn:
It's somewhat off-topic, but if you want to reduce symbols and template instantiations, a first easy step would be to replace the use of `std::conditional_t` with the implementation suggested in [this article][1] which basically amounts to this:

```cpp
template<bool>
struct conditional
{
    template<typename T, typename U>
    using type = T;
};

template<>
struct conditional<false>
{
    template<typename T, typename U>
    using type = U;
};

template<bool B, typename T, typename U>
using conditional_t = typename conditional<B>::template type<T, U>;
```

This implementation of conditional only produces template instantiations for `true` and `false` and everything else is aliases. According to the article it can even improve compile times. considering that `std::conditional` is used in a number of places in units.h, that would be an easy first step to mitigate the problem.

  [1]: https://odinthenerd.blogspot.com/2017/03/start-simple-with-conditional-why.html

johelegp:
I mean to reduce instantiations of types in this library, so no unnecessary work is done by design. I'd rather not reimplement `std::` stuff unless the amount of symbols produced is actually proven to be problematic, like someone hitting the 2^16 limit mentioned above and that `conditional_t` implementation reducing the symbols by 1/4th.

johelegp:
I have started adding some macros meant to resolve #133 and #138. I didn't edit the current macros so that I can test the new ones incrementally. If they work, I'll update the current ones accordingly and discard the new ones.

nholthaus:
I may have merged this badly...

johelegp:
This is a mess of rebases and manual merging conflict resolutions of all my concurrent PRs. Don't worry about it.

johelegp:
I have removed all unrelated commits. I'll now start trying out the macros.

nholthaus:
I'd be OK w/ clang 6 and gcc 7 as minimum versions (seems fair for c++17), but would really like to continue to support VS 2017.

johelegp:
`dimensionless` is now workable with CTAD. I've had to remove defaulting the `Underlying` template parameter of its definition as it interferes with CTAD. The only place where it could help is in a default constructor deduction guide, which I can add if requested.

nholthaus:
yeah to keep the syntax consistent (and to be consistent w/ previous usage) I think it would be good to include the deduction guide. As a design goal, we shouldn't require template usage in end-user code for basic dimensional-analysis functionality. 

I'm going to play with this over the weekend so I can get a feel for the changes. Looks good overall. Did you notice whether it helped with the error messages or not?

johelegp:
I have yet to check on the produced error messages.

johelegp:
I've rebased and cleaned up the commits. Seems like VS2015 didn't work with the deduction guide for the default constructor.

nholthaus:
yeah, renaming should be a PR on its own. I've been meaning to solicit some input from you about that. It was part of my initial motivation for 3.0 but I'm wondering if we're making a useful change or breaking compatibility for no good reason. I'm kind of on the fence.

johelegp:
VS2017's error seems to be SFINAE related. I hope I can decipher it.

johelegp:
Looks like the decibel units of integral underlying type are too error prone:
```C++
units/unitTests/main.cpp:1577: Failure
      Expected: ++b_dBW
      Which is: 2 dBW
To be equal to: dBW_t(2)
      Which is: 0 dBW
units/unitTests/main.cpp:1578: Failure
      Expected: b_dBW++
      Which is: 2 dBW
To be equal to: dBW_t(2)
      Which is: 0 dBW
units/unitTests/main.cpp:1579: Failure
      Expected: b_dBW
      Which is: 3 dBW
To be equal to: dBW_t(3)
      Which is: 0 dBW
units/unitTests/main.cpp:1580: Failure
      Expected: +b_dBW
      Which is: 3 dBW
To be equal to: dBW_t(3)
      Which is: 0 dBW
units/unitTests/main.cpp:1581: Failure
      Expected: b_dBW
      Which is: 3 dBW
To be equal to: dBW_t(3)
      Which is: 0 dBW
[  FAILED  ] UnitContainer.unitTypeUnaryAddition (0 ms)
units/unitTests/main.cpp:1685: Failure
      Expected: --b_dBW
      Which is: 3 dBW
To be equal to: dBW_t(3)
      Which is: 0 dBW
units/unitTests/main.cpp:1686: Failure
      Expected: b_dBW--
      Which is: 3 dBW
To be equal to: dBW_t(3)
      Which is: 0 dBW
units/unitTests/main.cpp:1687: Failure
      Expected: b_dBW
      Which is: 2 dBW
To be equal to: dBW_t(2)
      Which is: 0 dBW
units/unitTests/main.cpp:1688: Failure
      Expected: -b_dBW
      Which is: -2 dBW
To be equal to: dBW_t(-2)
      Which is: 0 dBW
units/unitTests/main.cpp:1689: Failure
      Expected: b_dBW
      Which is: 2 dBW
To be equal to: dBW_t(2)
      Which is: 0 dBW
[  FAILED  ] UnitContainer.unitTypeUnarySubtraction (1 ms)
```

johelegp:
```C++
/home/travis/build/nholthaus/units/unitTests/main.cpp:2476: Failure
      Expected: "8 cu_ft"
To be equal to: output.c_str()
      Which is: "0.226535 m^3"
```
This test is failing because `pow<>` doesn't preserve the strong type alias to the unit. This is currently only done in `std::common_type` in a few cases. I wonder how far we might want to go with this preservation.

johelegp:
Now I'm trying to solve https://github.com/nholthaus/units/pull/126#issuecomment-389393398 and 69bb2fb.

johelegp:
Another thing that could help with the error messages is making the `unit_conversion`s themselves strong type aliases. Even `units::unit<units::unit_conversion<... dim ... length ... std::ratio<0> ... std::ratio<0>, std::ratio<0>>, int>`, which got shorter thanks to variadic dimensions, is less readable than `units::unit<units::length::meter, int>`. That's not too far from the `units::length::meter_t<int>` which this PR would offer. Perhaps it might even be preferable in the face of possible CTAD language support for alias templates and it seeming like a less invasive change in terms of LoC and symbols.

Then we could invest on making the unit manipulators result in strong type aliases themselves, like having `units::divide<units::dimensionless_unit, units::time::seconds>` be that rather than `units::unit_conversion<... dim ... seconds ... std::ratio<-1> ... std::ratio<0>, std::ratio<0>>` if `units::frequency::hertz` didn't exist.

I think this alternative, yet nonconflicting approach is worth investigating. Do you mind if I open a separate PR for that?

nholthaus:
Yeah I think it's worthwhile. There was a point in the early 2.0's where
most of the (formerly named) units were still structs and we had error
messages similar to `units::unit<units::length::meter, int> `. They were
very clear except for complex ephemeral units (which we don't need to care
about so much since they're basically user defined and domain specific, or
intermediate). No one was really complaining about complex errors back then.

Am I wrong in saying that this approach is not only non-conflicting, but
actually complementary to the current PR? As far as I can tell, everything
we strengthen is a net win.

johelegp:
Yeah, they can also be complementary.

johelegp:
I've been thinking of splitting these changes in at least three, in no particular order:

1. Making the `_t` suffixed units parametrized over the underlying type.
2. Making the `_t` suffixed units strong type aliases.
3. Adding explicit support for CTAD.

By separating 2. and 3., the tests would have to explicitly state the underlying type template argument. That prevents running into the CTAD bugs of compilers with incomplete C++17 support. Then, we could add tests for explicitly testing CTAD.

johelegp:
The reason for not using CTAD in the tests not testing CTAD, as mentioned above, is to avoid contributors having to consider when they're able to use it, since the tests they add and edit will go through compilers with incomplete C++17 support. That's fine, since even today, the tests aren't written like a user would. Indeed, many places use `meter_t(1)` instead of the more natural `1_m` (which I'd expect would be more used than the former).

nholthaus:
That's fine w/ me. The tests were originally written with VS 2013 support
in mind. As long as the coverage is good, I don't think it matters.

Probably a good idea to include a CTAD test wrapped in an `#ifdef` which
can be enabled from cmake.

nholthaus:
I was looking back over the comments here and can't really tell what state the PR is in. Once we pull in the strong conversions, we should chat about what if anything is still outstanding.

Once we get this in and the renaming in, we're only some docs and a couple small issues away from a 3.0 build I think.

johelegp:
The state is that it does too many things, so I've been splitting them into other PRs. Soon, it should only be about making the units strong type aliases, as the title suggests.
>Once we get this in and the renaming in, we're only some docs and a couple small issues away from a 3.0 build I think.

I agree with that.

johelegp:
I can't catch what VS2017 is complaining about.

Anyways, all that's missing is testing CTAD and maybe doing something about the ODR issues.

johelegp:
I've thought about what could be done about the ODR issues.

1. Forward declare unit class templates and define `strong` specializations in `<units/core.h>`, like
```C++
template<class Underlying>
class square_meter_t;
namespace traits
{
    template<class Underlying>
    struct strong<conversion_factor<std::ratio<1>, units::dimension::area>>
    {
        using type = square_meter_t<Underlying>;
    };
}
```
Then, a user who `#include`s `<units/length.h>` and needs the return type of `1_m * 1_m` to be complete gets an error saying that `square_meter_t` is undefined unless they also `#include`d `<units/area.h>`.
2. Document preferring `#include`ing `<units.h>` over the finer-grained headers, and/or warn about the ODR issues when using them.

If doing 1., some compilation time would be lost compared to number 2, but it shouldn't be much compared to actually defining them. It shouldn't be time-consuming to implement, either: Take a snapshot of the macros defining units before they were scattered in different headers, apply some regexes to change, for example,
```C++
    UNIT_ADD(area, square_meter, square_meters, sq_m, conversion_factor<std::ratio<1>, units::dimension::area>)
```
to
```C++
    UNIT_DECLARE(area, square_meter, conversion_factor<std::ratio<1>, units::dimension::area>)
```
which becomes the first piece of code above, and finally browse the commits looking for renames and new units, and manually apply this new macro to them.

How do you think we should move forward?

unitTests/main.cpp
```diff
@@ -1018,64 +1065,246 @@ TEST_F(UnitContainer, constructionFromUnitContainer)
...
 
 TEST_F(UnitContainer, assignmentFromUnitContainer)
 {
-	const unit<meters, int> a_m(1);
-
-	unit<meters, int> b_m;
-	b_m = a_m;
-	EXPECT_EQ(1, b_m());
+	unit<meters, int> a_m(1);
+	a_m = +a_m;
```
johelegp: The `+` in these `self = +self` is to avoid warnings on self-assignment.

johelegp:
This PR is ready.

include/units/core.h
```diff
@@ -165,10 +165,91 @@ namespace units
...
+			return ::units::unit_name_v<Unit>; \
+		} \
+\
+		template<class Unit = unitName> \
+		constexpr const char* abbreviation() const noexcept \
+		{ \
+			return ::units::unit_abbreviation_v<Unit>; \
+		} \
+	}; \
+\
```
nholthaus:
```cpp
\
	unitName()->unitName<UNIT_LIB_DEFAULT_TYPE>; \
```

I want a template deduction guide for the default constructor that uses the libraries default type. This will provide backwards compatibility to 2.3 code (yes I'm ignoring the fact that we're going to change all the names too), and also it keeps things simple for the template-averse or in situations where the underlying type isn't very important.

johelegp: Very well. I'll first rebase and then add a commit with that. IIRC from previous tries:
- That'll require braces, like in `meter_t{}`, as `meter_t` is ill-formed.
- Adding that to the `Underlying` class template parameter made all units have `double` underlying type, so the solution above would still require using `meter_t<double>` for function arguments and such.

nholthaus:
It looked OK in a quick-and-dirty test but I may not have run into those
situations. Can you provide examples before we decide?

johelegp:
You mean about this: https://github.com/nholthaus/units/pull/150#discussion_r228650793? I might be able to excavate some refs.

johelegp:
Done.

nholthaus:
looks like the brackets for the default constructors are only necessary for `const` instantiations... because a `const` must be initialized on creation. Non-const default constructed units don't need brackets, as expected. I'm going to look at function declarations next to see if there's any weirdness to that.

johelegp:
Oh, of course. Thank you.

johelegp:
Why a0bdbf5? They were supposed to test the default construction of `const` qualified units through CTAD.

PR #154, opened 2018-07-08
Remove unused stuff

johelegp:


nholthaus:
two questions:
- Can you remind me how/why we don't need `has_same_scale` anymore?
- do we have a trait to test whether a type `T` is a dimension or not?

johelegp:
>Can you remind me how/why we don't need `has_same_scale` anymore?

`has_same_scale` predates mixed-underlying type support. Using it constrains the `nls` member, and thus the `underlying_type` member, to being the same type.

>do we have a trait to test whether a type `T` is a dimension or not? (I think it will be handy for concepts/user defined dimensions)

No. This one seems a leftover of the previous `base_unit`, which hard-coded all dimensions. It's unused in the current implementation.

Maybe I'm rushing it with these removals. Please, close it if you think so.

nholthaus:
No, I'm just trying to get on the same page. Anything removed can always
be put back, and my guess its if we did need them it would be in a newer
form based on what you said.

include/units.h
```diff
@@ -2703,7 +2703,6 @@ namespace units
 
 		template<typename... T>
 		inline constexpr bool has_decibel_scale_v = has_decibel_scale<T...>::value;
-	}
```
johelegp: This `}` should be there to close the namespace. Was there a conflict with v3.x that prevented fast-forwarding? Otherwise, I think rebasing is the better option, as it marks conflicts like these to be solved.

nholthaus:
I've used github's web interface for conflict resolution, but it's not very
good. Back to the command line!

PR #163, opened 2018-07-19
Fix streaming of ephemeral units of integer underlying type

johelegp:
When the conversion to the base unit is lossy, it is a compile-time error.
This solution might incur accuracy loss as the unit is
converted to a base unit of `double` underlying type.

Any idea on how to avoid the accuracy loss?

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -2422,7 +2422,7 @@ namespace units
 	{
 		using BaseUnit =
 			unit_conversion<std::ratio<1>, typename traits::unit_conversion_traits<UnitConversion>::dimension_type>;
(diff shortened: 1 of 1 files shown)


nholthaus:
Ideally lossy conversions would result in warnings rather than compile-time errors (mimicking existing c++ behavior). Nothing obvious springs to mind, but I'll play with it.

johelegp:
The streaming operator for ephemeral units streams [coherent derived unit](https://www.bipm.org/en/publications/si-brochure/section1-4.html).

Printing an object of type `unit<millimeter, int>` fails to compile because converting to its coherent form of type `unit<meter, int>` is lossy.

PR #167, opened 2018-07-26
`unit_conversion`s as strong type aliases

### LINKED ISSUE #160: No more nested namespaces in `units`
Following https://wg21.link/p0816 and common design guidelines, I suggest the following.

Mark the namespaces in which the units are defined as `inline`. The README suggests the namespaces are used to avoid name clashing, for example, `mass::pounds` and `force::pounds`. By marking the namespaces as `inline`, these rare name clashes are avoided, as attempting to use something like `units::pounds` becomes a compile-time error. We also benefit from using most units by simply prepending them with `units::` and avoiding the pitfalls used to circumvent nested namespace designs. These names might be noise in their context of use, like the `length` in `units::length::meters`, or plain overly long, like `units::magnetic_field_strength::tesla`, to pick one. The user is still free to use them if they so decide.

johelegp:
Resolves #160 to prevent having to prefix the `unit_conversion` tags everywhere so that they're valid outside their namespace [here](https://github.com/nholthaus/units/compare/v3.x...johelegp:strong_type_alias?expand=1#diff-cc75003c399d878c360873c2c8c783d6R152).

The new type trait, `strong`, maps a `unit_conversion` to its strong type alias. They are used as late as possible to prevent ODR issues. Manipulators like `squared` don't use it as they're used in the definition of units. See, for example:
```C++
UNIT_ADD(solid_angle, degree_squared, degrees_squared, sq_deg, squared<angle::degrees>)
```
This declares `degrees_squared` as a strong type alias of `squared<angle::degrees>`, and then specializes `strong` so that `strong_t<squared<angle::degrees>>` names `degrees_squared`. If `squared` used `strong_t`, then the meaning of `squared<angle::degrees>` would change after this unit definition.

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -102,8 +102,10 @@ namespace units
 		 */
 		namespace detail
 		{
-			using PI = conversion_factor<std::ratio<1>, dimensionless_unit, std::ratio<1>>;
-		}
+			struct PI : conversion_factor<std::ratio<1>, dimensionless_unit, std::ratio<1>>
+			{
+			};
+		} // namespace detail
 
 		// clang-format off
 		inline constexpr const unit<detail::PI>																												pi(1);											///< Ratio of a circle's circumference to its diameter.
--- a/include/units/core.h
+++ b/include/units/core.h
@@ -158,7 +158,7 @@ namespace units
  * @param		nameSingular singular version of the unit name, e.g. 'meter'
  */
 #define UNIT_ADD_UNIT_DEFINITION(namespaceName, nameSingular) \
-	namespace namespaceName \
+	inline namespace namespaceName \
 	{ \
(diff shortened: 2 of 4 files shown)


johelegp:
Here's a [diff of errors](https://www.diffchecker.com/JdY5Y8c4) based on [these Compiler Explorer sources](https://godbolt.org/g/mjMPtD). It seems that the lack of strong typing can pull some overloads from elsewhere.

nholthaus:
looks great. The improvement to the error messages is phenomenal. I pulled the header split PR first so I think once the conflicts are resolved this is ready to go.

johelegp:
Rebased. clang-format was ran on the `conversion_factor` renaming.

johelegp:
I meant to say that the code was auto formatted for me. You should turn it on.

nholthaus:
roger that.

PR #171, opened 2018-08-09
Split headers

### LINKED ISSUE #164: Splitting unit definitions into different headers
Related issues/comments:
- #48
- #77
- https://github.com/nholthaus/units/issues/160#issuecomment-406042056

Regarding https://github.com/nholthaus/units/issues/48#issuecomment-329987514
>deprecated by the define/cmake options to opt-in to specific namespaces

By splitting and not relying on those macros, there's less chances of clash between depending projects using `units`.

johelegp:
Resolves #164.

How does it look like so far? If it's good, I'll continue with the rest of dimensions.


johelegp:
I tried to rebase the first commit, by first doing a `git mv include/units.h include/units/core.h` and then moving non-core functionality into `units.h`, in order to reduce the number of apparent changed lines and ease review. It failed, and worse, the changed lines count increased.

include/units/angle.h
```diff
@@ -0,0 +1,306 @@
...
+#endif
+	/**
+	 * @ingroup		UnitMath
+	 * @brief		Compute tangent
+	 * @details		The input value can be in any unit of angle, including radians or degrees.
+	 * @tparam		AngleUnit	any `unit` type of `dimension::angle`.
+	 * @param[in]	angle		angle to compute the tangent of
+	 * @returns		Returns the tangent of <i>angle</i>
+	 */
+#if !defined(DISABLE_PREDEFINED_UNITS) || defined(ENABLE_PREDEFINED_ANGLE_UNITS)
```
nholthaus: These definitions/macros should go away. They're redundant with split headers.

nholthaus:
@johelegp per your earlier comment... yikes. We can't guarantee (and don't want to) that a definition will exist for every ephemeral unit type. That sounds like a show-stopper for strong type aliases.

johelegp:
As explained above (https://github.com/nholthaus/units/pull/171#issuecomment-411916790), the reason it would result in a compile-time error is because `core.h` would include the forward declaration of `square_meter_t` and specialization of `traits::strong` or similar for it, which the operator would use for its return type. Ephemeral units don't need that.

include/units/temperature.h
```diff
@@ -0,0 +1,75 @@
...
+#pragma once
+
+#ifndef units_temperature_h__
+#define units_temperature_h__
+
+#include <units/core.h>
+
+namespace units
+{
+	// NOTE: temperature units have special conversion overloads, since they
```
johelegp: Is this still relevant?

PR #193, opened 2018-10-08
Simplify `unit::toLinearized`

johelegp:
By defaulting `units::toLinearized`'s return type to the unit's `underlying_type`, call sites that don't need something else get simpler.

What do you think about simplifying it further, so that it's not a template function? Changing `x.template toLinearized<Arithmetic>()` for `static_cast<Arithmetic>(x.toLinearized())` is, I believe, more clear on the intention and easier to read at call site.

### DIFF
--- a/include/units/core.h
+++ b/include/units/core.h
@@ -1816,9 +1816,8 @@ namespace units
 	template<class UnitTo, class UnitFrom, class = std::enable_if_t<detail::is_convertible_unit<UnitFrom, UnitTo>>>
 	constexpr UnitTo convert(const UnitFrom& from) noexcept
 	{
-		return UnitTo(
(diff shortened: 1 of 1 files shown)


nholthaus:
yeah, removing the template would be more idiomatic.

johelegp:
> yeah, removing the template would be more idiomatic.

Am I correct to assume that you agree with the "further simplification" as posted in the second part of the OP? I'm asking because I was going to implement that.

The commit for (manually) closing this PR looks wrong and doesn't include a part of the changes, presumably due to the conflicts (for having added `linearized_value`).

nholthaus:
yes agree to the second part. go ahead and deconflict as much as you can,
especially #150.

PR #198, opened 2018-10-13
Fix data transfer rate unit's dimension

johelegp:
I found it weird when I first read its declaration, as I'd expect it to be `dimension_divide<data, time>` and not `make_dimension<data, std::ratio<-1>>` like `frequency` is to `time`. Earlier, I finally tried to mix a data transfer unit with a time unit, and chaos ensued.


nholthaus:
good catch... that made no sense on my part.

Issue #185, opened 2018-09-19
Lying traits


JohelEGP:
`is_convertible_conversion_factor` is not about testing whether a `conversion_factor` is convertible to another, but whether their dimensions are the same, as suggested by [its implementation](https://github.com/nholthaus/units/blob/5483589322daad285eedd4e2822058ba4e6640d8/include/units/core.h#L1571). Maybe it's its name that needs to be improved.

The traits with a `type` member alias shouldn't default to `void`, but to having no `type` member, like `std::enable_if`. Remember that `static_assert` that never failed because it eventually indirectly evaluated `std::is_same_v<void, void>`?

nholthaus:
very clever issue name my friend. You're right, and I think that our 4 or 5 time replace all on the name `unit` has led us to the right name for the trait... how do you feel about `is_same_dimension`?

The second paragraph I need to prototype. IIRC the reason for defaulting to void was that compilation would fail when using the dimension traits if the two types weren't of the same dimension, but hopefully we've moved past that now.

Guillaume227:
There is also a lying template name, if I may :)

`NonLinearScale optional scale class for the units. Defaults to linear`
I think it's confusing to have a NonLinear concept encompass a linear variety.
Calling it something like 'NumericalScale would be clearer.

johelegp:
An example of the `void` issue is [`detail::is_convertible_conversion_factor`](https://github.com/nholthaus/units/blob/5483589322daad285eedd4e2822058ba4e6640d8/include/units/core.h#L1661), which adds logic to [`is_convertible_conversion_factor`](https://github.com/nholthaus/units/blob/5483589322daad285eedd4e2822058ba4e6640d8/include/units/core.h#L1571). It needs additional checks in case the trait's result would lie so that it can successfully used in SFINAE contexts.

johelegp:
I think `is_nonlinear_scale` is a lying trait, too, based on how `unit` uses it. It only tests for default constructability, but the tested type is constructed from the underlying type.

I think the classes that implement this concept can be simplified and improved by dropping the `Args` variadic template argument of their constructor and instead using a `store_linear_value` tag. `linear_scale` can default it and `decibel_scale` overload based on it. Then we can rid ourselves of those `std::true_type() /*store linear value*/` lines and use `store_linear_value()`.

johelegp:
[...]

PR #199, opened 2018-10-20
Simplify numerical scales

johelegp:
- Change term of non-linear scale to numerical scale. Suggested in https://github.com/nholthaus/units/issues/185#issuecomment-423107386.
- Add `linearized_value_t` tag to disambiguate the `unit` constructor whose value is already linearized.
- Reduce the scales' scope of functionality to what their name suggest. Make them non-template entities with only a `linearize` and `scale` static member functions in place of their old constructor and `operator()` respectively. Move `m_value` and `operator()` to `unit`, where they naturally belong.
- `unit` still inherits from `NumericalScale`, but privately. This is so that the [scale traits](https://github.com/nholthaus/units/blob/5483589322daad285eedd4e2822058ba4e6640d8/include/units/core.h#L2718-L2754) continue to work.
- Rename `m_value` to `linearized_value`, upgrading its explaining comment to code. [\[1\]](https://github.com/nholthaus/units/blob/5483589322daad285eedd4e2822058ba4e6640d8/include/units/core.h#L2794) [\[2\]](https://github.com/nholthaus/units/blob/5483589322daad285eedd4e2822058ba4e6640d8/include/units/core.h#L3314)
- Drop `constexpr` support from `decibel_scale`, which uses non-`constexpr` `std` math functions.
- Remove `has_value_member`, now unused.

With this, the fact that the unit values are stored linearized has effectively become an implementation detail, I think.

I think this is desirable as the current scales can be abused because they're publicly bases of `unit`, their constructors were overly accepting, and the non-linear scale concept is underspecified.

### DIFF
--- a/include/units/core.h
+++ b/include/units/core.h
@@ -2119,8 +2119,8 @@ namespace units
 		 * @details		constructs a new unit with `value`.
 		 * @param[in]	value	unit magnitude.
 		 */
-		template<class Ty,
-			std::enable_if_t<!traits::is_dimensionless_unit<UnitType>::value && detail::is_non_lossy_convertible<Ty, T>,
+		template<class Ty, class Cf = UnitType,
+			std::enable_if_t<!traits::is_dimensionless_unit<Cf>::value && detail::is_non_lossy_convertible<Ty, T>,
 				int> = 0>
 		explicit constexpr unit(const Ty value) noexcept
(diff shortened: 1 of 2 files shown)


nholthaus:
great change. I had in my head at one point that the scales could support really whacky things (like richter scale or other offset scales similar to how temperature is handled), but it was very poorly named and pointlessly complex in the end, in addition to never being exercised.

PR #181, opened 2018-09-15
V3.x

nanowizard:
Adds substance_mass.h and substance_concentration.h and adds metric prefixes to substance units.

PR #201, opened 2018-10-26
Fix substance units recently added in #181

johelegp:
Please, review @nholthaus and specially @nanowizard. I added some quick, short tests.


nholthaus:
thanks, I'll add in some more tests

nanowizard:
Looks good to me. Thank you!

Issue #196, opened 2018-10-12
Better use of abbreviated unit identifiers


JohelEGP:
For a long while now, I've thought that it'd be great if we could naturally write the following.
```C++
using namespace units::literals;
using namespace units::constants;
const auto speed_limit = 90_km/h;
```
However, we need to write `90_km/1_h` as the abbreviated unit identifiers happened to be taken up by some "tags."

What do you think about repurposing these identifiers to somewhere else, like the `units::constants` namespace, where they'd be declared along the lines of:
```C++
inline constexpr time::hour_t<int> h(1);
```
This would allow us to more naturally express units in our programs. The standard time library already allows expressing dates using the natural syntax, like [`1970y/January/1`](http://eel.is/c++draft/time#clock.system.overview-1.example-1). I don't see why we shouldn't follow suit.

nholthaus:
Yes, this. It never crossed my mind honestly but I _love_ the natural syntax. That's the ethos and purpose of the library in the first place.

johelegp:
I had thought about it, and people `using namespace units::literals;` for the UDLs might find themselves having undesired clashes or accidentally using these kind of "constant literals". However, I haven't come up with something better.

nholthaus:
hmm, good point. I still like inlining the real constants namespace, we can
call these `literal_constants` or something like that. That said, I may be
OK leaving the clash resolution as an exercise to the user. If you have a
bunch of single letter variables or unit abbreviations as variables, you
may need/want a refactor for units anyway.

johelegp:
Yeah. `literals` might be good enough most of the time.

nholthaus:
if not we'll just patch it out in v3.1 I suppose...

johelegp:
#256 removes the abbreviation-named tags.

johelegp:
#212 blocks a general solution for this.

nholthaus:
all the concentration unit names were expanded, so only dB uses an abbreviation. I'm pretty sure we don't want to use operators like `/` and `*` to change linear scales, since arithmetic has a different meaning in log space.

johelegp:
Alternatively, `dB` should be pluralized, like how all other units were renamed (I don't know if there are others). Then `dB` would be freed up for this.

nholthaus:
I don't have a problem with pluralizing the names, but I don't think it will work as intended. I can't do `20 * dB * W` because `operator*` isn't defined for `decibel_scale` types, and I think even if it did that would still result in `watts`?

[...]

Issue #159, opened 2018-07-18
Renaming the units to their natural spelling


JohelEGP:
For example, renaming `meter_t` to `meters`. This has probably been discussed in various places. I'll link to them as I find them.

To answer https://github.com/nholthaus/units/pull/150#issuecomment-405940556, I believe that, in the long run, renaming the units to their natural spelling is the right way to go. The `_t` suffix already has some uses that don't match the use in this library's units, like in the the standard type traits and the arguably more similar use in the types of `<cstdint>`. As the library becomes more popular, the `_t` suffix will become more of a baggage than the natural spelling the library can exploit with its CTAD capabilities and further possible improvements.

nholthaus:
I agree with the logic. v3.0 is going to have a hefty upgrade burden on the users from 2.3, but that's the price of progress. We'll need an upgrade guide in the docs.

johelegp:
What should we do about
https://github.com/nholthaus/units/blob/5483589322daad285eedd4e2822058ba4e6640d8/include/units/pressure.h#L75
and similar cases which have the same singular and plural name?

johelegp:
[`dimensionless_unit`] also needs a name revision.

nholthaus:
Done in 3.1.0: units use their natural plural spelling (`meters`); CTAD makes the angle brackets optional. Closing.

PR #202, opened 2018-10-26
inline namespace constants

nholthaus:
in the spirit of our other namespace inlining, I'm inlining the constants namespace as well

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -87,6 +87,14 @@
 
 namespace units
 {
+
+	namespace detail
+	{
+		struct PI : conversion_factor<std::ratio<1>, dimensionless_unit, std::ratio<1>>
+		{
+		};
+	} // namespace detail
+
 	//------------------------------
 	//	CONSTANTS
 	//------------------------------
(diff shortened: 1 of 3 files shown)


johelegp:
I hadn't suggested it due to constants like `h`, which clash with their equivalently named `conversion_factor`, due to my desire now expressed in #196, and future constants possibly having the same problem. Not that I suggest using the `conversion_factor`s with the abbreviated name.

nholthaus:
yeah I think we should just change the names of the factors right?

johelegp:
What do you mean? Right now, we have `m`, `meter` and `meters`, which are `conversion_factor`s, and the unit `meter_t`. Following the units' renaming, there would be `m` and `meter`, and `meters`, respectively. There's also the issue of https://github.com/nholthaus/units/issues/159#issuecomment-427608557.

nholthaus:
@johelegp once I clear the PR queue, I'm thinking of ding the unit rename, e,g. removing the `_t`. To do that, I'll first have to change the conversion factor names. I'm thinking unit_name_singular + `_conversion`. What do you think?

Once that's done, the abbreviations will be free again and we can do the `h(1)` defines, as well as define the SI unit of each dimension in `core.h`

PR #206, opened 2018-11-02
Update README [skip ci]

johelegp:
There's room for improvement.


johelegp:
A few changes are about how there are `_t` units and plural-named template units, and how there's just not a single header anymore, so it might be worth adding sections about those apart from the in-place fixes I did.

PR #207, opened 2018-11-03
Fix common type of dimensionless and arithmetic types

johelegp:
I didn't consider these specializations back then. While further testing [my quantity concepts](https://github.com/johelegp/Made_in_Abyss/blob/master/doc/concepts.md#quantities-conceptsobjectqty), I tested whether the dimensionless units and arithmetic types behaved like quantities of the same dimension, and it failed. I don't know if it's the cause, but I realized that there weren't `std::common_type` specializations for dimensionless units and arithmetic types, unlike how there are ordering and arithmetic operators for them. You'll see that, in the test I added in this PR, there are inconsistencies in the common type. Some are dimensionless units, other are arithmetic types.

### DIFF
--- a/include/units/core.h
+++ b/include/units/core.h
@@ -390,10 +390,10 @@ namespace units
 			using type = globalUnitName<common_type_t<Underlying1, Underlying2>>; \
 		}; \
 \
-		template<typename UnderlyingLhs, template<class> class StrongUnit, class UnderlyingRhs> \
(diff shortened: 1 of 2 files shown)


nholthaus:
looks like it doesn't work on gcc, but it's not immediately clear why. May be a bug since the most pedantic compiler (clang) and least pedantic (msvc) both build it OK.

johelegp:
You'll notice that the added tests are commented out. That's because I was working towards making them valid, and the GCC hiccup was more about a drive-by fix.

PR #208, opened 2018-11-03
Teach `is_`x`_unit` to be a `UnaryTypeTrait`

johelegp:
For the longest while, I've been trying to fix and improve the type traits. Since success eludes me, I'm going to push what I can. Slowly, but eventually, that should improve the situation.

[`UnaryTypeTrait`](https://timsong-cpp.github.io/cppwp/n4659/meta.rqmts#def:UnaryTypeTrait).

### DIFF
--- a/include/units/core.h
+++ b/include/units/core.h
@@ -508,19 +508,19 @@ namespace units
 
 #define UNIT_ADD_DIMENSION_TRAIT(unitdimension) \
 	/** @ingroup	TypeTraits*/ \
-	/** @brief		Trait which tests whether a type represents a unit of unitdimension*/ \
-	/** @details	Inherits from `std::true_type` or `std::false_type`. Use `is_ ## unitdimension ## _unit_v<T>` to \
-	 ** 			test the unit represents a unitdimension quantity.*/ \
-	/** @tparam		T	one or more types to test*/ \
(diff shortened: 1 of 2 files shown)


nholthaus:
I'm not so sure about 23eba91. When used with a single argument, the traits operate semantically equivalently to an ISO unary trait. Allowing multiple arguments just improves the readability when you want to check multiple types, but in my opinion the semantics are still equally clear. I think the standard is unnecessarily limiting in this case, and wasn't written with variadic templates in mind.

That said, I'm not necessarily against it and I'd like to understand your rationale. Also which parts of the traits did you fail to improve? I know there's a lot and some are pretty yucky (sqrt...)

johelegp:
I know it's convenient, but I made the traits non-variadic because the [additional arguments are supposed to help define the property being described](http://eel.is/c++draft/meta#rqmts-1.sentence-2), as the `std::is_`(`nothrow_`|`trivially_`)(`move_`|`copy_`|`default_`)`constructible` family of type traits does. Just me being pedantic. I can revert that part.

About the other traits improvements, I've mentioned some before. There's the problematic defaulting to `void`, the `traits::is_unit_v` vs `detail::is_unit` because the former isn't enough, and probably some others.

PR #211, opened 2018-11-05
V3.x rename conversion factors

nholthaus:


nholthaus:
@johelegp meant to ask you about this:

```
[  FAILED  ] TypeTraits.is_data_unit (0 ms)
[ RUN      ] TypeTraits.is_data_transfer_rate_unit
/home/travis/build/nholthaus/units/unitTests/main.cpp:603: Failure
Value of: (traits::is_data_transfer_rate_unit_v<const gigabytes_per_second<double>&>)
  Actual: false
Expected: true
```

Why don't we want const references to pass the dimension trait?

johelegp:
For the same reason we don't want it working on pointers. They're pointer/references to types, not the types themselves.

nholthaus:
I feel like this is one of those "c++ defaults are backwards" situations. The reason for the existence of dimension traits is for sfinae/concepts, where you may save a conversion over defaulting to common_type, etc. So all I really want to know is, "is this the dimension of unit that I want". 99% of the time const/ref/ptr doesn't matter for the concept to be valid, so it seems better to check that in addition the 1% of the time than have to wrap most calls to the trait with `remove_ref`.

johelegp:
That's alright, but it becomes a gotcha where it appears to work like the standard's design. Note all the other occurrences of tests in #208 that need to be updated after 484a58b.

include/units/core.h
```diff
@@ -2508,8 +2511,8 @@ namespace units
 			std::enable_if_t<detail::is_time_conversion_factor<Cf> && detail::is_losslessly_convertible<Rep, T>, int> =
 				0>
 		constexpr unit(const std::chrono::duration<Rep, Period>& value) noexcept
-		  : linearized_value(NumericalScale::linearize(units::convert<unit>(
-				units::unit<units::conversion_factor<Period, dimension::time>, Rep>(value.count()))()))
+		  : linearized_value(NumericalScale::linearize(units::convert<unit>(units::unit<units::conversion_factor<Period, dimension::time>, Rep>(value.count()))
```
johelegp: This line is longer than the Clang Format limit. Why don't you turn it on to run on every save?

nholthaus: it's on, I think visual studio's implemention isn't very good. I've seen it crap out a few times.

johelegp: I've been looking into automating it in a Travis build. I'll get back to you when I'm successful.

nholthaus:
fixes #159

include/units/core.h
```diff
@@ -510,14 +513,14 @@ namespace units
 #define UNIT_ADD_DIMENSION_TRAIT(unitdimension) \
 	/** @ingroup	TypeTraits*/ \
 	/** @brief		`UnaryTypeTrait` for querying whether `T` represents a unit of unitdimension*/ \
-	/** @details	The base characteristic is a specialization of the template `std::bool_constant`. \
+	/** @details	The base characteristic is a specialization of the template `std::bool_constant`. \ \ \ \ \
```
johelegp: I see what you mean WRT MSVC's Clang Fromat.

johelegp:
> fixes #159

I'm not sure if it works with you even though it's disabled, but that needs to go in the first comment.

That said, aren't you renaming the units to their plural spelling?

nholthaus:
no I was planning on using the singular form. Do you have a preference?

johelegp:
Well, I thought the natural form in english was the plural spelling. Just like `std::chrono::seconds`.

nholthaus:
yeah good point. It's just a more difficult refactor, but it shouldn't be a problem

unitTests/main.cpp
```diff
@@ -180,622 +181,501 @@ TEST_F(TypeTraits, replace_underlying)
...
+	static_assert(
+		has_equivalent_conversion_factor(std::common_type_t<kilometer<int>, millimeter<int>>(), millimeter<int>()));
+	static_assert(
+		std::is_same_v<std::common_type_t<meter<int>, kilometer<int>>, std::common_type_t<kilometer<int>, meter<int>>>);
+	static_assert(std::is_same_v<std::common_type_t<meter<int>, millimeter<int>>,
+		std::common_type_t<millimeter<int>, meter<int>>>);
+	static_assert(std::is_same_v<std::common_type_t<millimeter<int>, kilometer<int>>,
+		std::common_type_t<kilometer<int>, millimeter<int>>>);
 
-	static_assert(has_equivalent_conversion_factor(
```
johelegp: I hadn't updated these tests to test `unit` too, and not only the strong types.

unitTests/main.cpp
```diff
@@ -862,27 +742,29 @@ TEST_F(STDTypeTraits, std_common_type)
...
-	EXPECT_EQ((std::hash<unit<kilometers, int>>()(unit<kilometers, int>(42))), 42);
+	EXPECT_EQ((std::hash<meter<int>>()(meter<int>(42))), 42);
+	EXPECT_EQ((std::hash<millimeter<int>>()(meter<int>(42))), 42000);
+	EXPECT_EQ((std::hash<millimeter<int>>()(millimeter<int>(42))), 42);
+	EXPECT_EQ((std::hash<kilometer<int>>()(kilometer<int>(42))), 42);
 
 	EXPECT_EQ((std::hash<dimensionless<double>>()(3.14)), std::hash<double>()(3.14));
-	EXPECT_EQ((std::hash<unit<dimensionless_unit, double>>()(3.14)), std::hash<double>()(3.14));
-	EXPECT_EQ((std::hash<dimensionless<int>>()(42)), (std::hash<unit<dimensionless_unit, int>>()(42)));
+	EXPECT_EQ((std::hash<dimensionless<double>>()(3.14)), std::hash<double>()(3.14));
```
johelegp: Same here. You'll see that the line above is now equal to this one. And below, too, at the tests for the ordering operators.

unitTests/main.cpp
```diff
@@ -1997,62 +1884,62 @@ TEST_F(UnitType, unitTypeMixedUnitMultiplication)
...
-	foot_t<double> a_ft(3.28084);
-	second_t<double> a_sec(10.0);
-	const unit<meters, int> d_m(1), e_m(2);
-	const std::common_type_t<unit<meters, int>, unit<feet, int>> j(d_m);
-	const unit<seconds, int> b_sec(10);
+	meter<double> a_m(1.0), b_m(2.0);
+	foot<double> a_ft(3.28084);
+	second<double> a_sec(10.0);
+	const meter<int> d_m(1), e_m(2);
+	const std::common_type_t<meter<int>, foot<int>> j(d_m);
```
johelegp: This is fine since the common type of `unit`s are tested above, but generally I don't think there were undesirable uses of `unit` left that weren't testing `unit` and not the strong types.

nholthaus: should I revert c8d935f, or would you rather patch up the tests?

johelegp: Reverting seems the better option.

include/units/core.h
```diff
@@ -2437,7 +2440,7 @@ namespace units
 	 *				- \ref constantUnits "constant units"
 	 */
 	template<class ConversionFactor, typename T = UNIT_LIB_DEFAULT_TYPE, class NumericalScale = linear_scale>
-	class unit : NumericalScale, units::detail::_unit
+	class unit : public ConversionFactor, private NumericalScale, units::detail::_unit
```
johelegp: This `private` is redundant.

johelegp:
I don't think I have anything else to share. Reviewing this allowed me to notice a few things, as stated in the new issues.

nholthaus:
I've been looking at the revert and actually I don't think are probably too many more places than what you pointed out where we shouldn't be using strong unit types. I like to keep the tests close to the intended usage of the syntax, and have 'bootstrapped' them a few times. We messed up the meanings of a few tests, but I'm fully confident that we didn't break any functionality.

let's fix the few we need to after the name refactor.

johelegp:
Note that if you change the name of the units to their plural spelling, parts of my opinion in #212 will be applied due to how the macros are used in `<units/concentration.h>`. And that the singular name part of the `UNIT_ADD` macro will become essentially unused, I believe.

nholthaus:
@johelegp the latest push fails one of the strong tests, but I'm not sure if something is wrong, or if the test is no longer valid because of how we define unit conversions in terms of other units in this branch. May want to take a look at the refactored `METRIC_PREFIXES` macro.

nholthaus:
@johelegp I re-tried this branch with the gcc 10.1 and the strong_t alias template still gives us trouble. Could we refactor this in such way that `strong_t` is also a struct instead of an alias template?

johelegp:
I'm back online. I'll try compiling it with GCC 10.2. Can you give me a github reference as to what's wrong with `strong_t`? Besides my issue to revert it. It's been a long while, after all, so I need the context.

johelegp:
GCC 10.2 compiled it successfully and only a locale test failed:
```
...
1: /home/johel/Documents/C++/Forks/units/unitTests/main.cpp:2524: Failure
1:       Expected: "de_DE.utf8"
1: To be equal to: setlocale(6, "de_DE.utf8")
1:       Which is: NULL
1: /home/johel/Documents/C++/Forks/units/unitTests/main.cpp:2529: Failure
1:       Expected: point_de
1:       Which is: '.' (46, 0x2E)
1: To be equal to: ','
1:       Which is: ',' (44, 0x2C)
1: /home/johel/Documents/C++/Forks/units/unitTests/main.cpp:2535: Failure
1:       Expected: "2,5 km"
1: To be equal to: units::length::to_string(de).c_str()
1:       Which is: "2.5 km"
...
1: [  FAILED  ] UnitType.to_string_locale
...
```

nholthaus:
sorry, I think I figured it out looking at the generated macro code, things
were confusing being in the middle of the rename.

We should figure out if we actually need to revert strong types or not
though. I don't really remember what we thought was wrong, and things seem
to be working on the surface.

nholthaus:
that's fine, the locale just isn't installed

PR #214, opened 2018-11-06
Fix `is_convertible_unit` and `is_same_dimension`

johelegp:
Partially addresses #185. The changed tests from the first commit were suffering from the `std::is_same_v<void, void>` syndrome.

### DIFF
--- a/include/units/core.h
+++ b/include/units/core.h
@@ -1880,23 +1880,24 @@ namespace units
 		/**
 		 * @ingroup		TypeTraits
 		 * @brief		Trait which checks whether two conversion factors have the same dimension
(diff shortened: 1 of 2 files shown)


include/units/core.h
```diff
@@ -2110,19 +2101,19 @@ namespace units
 		 *				`conversion_factor` (-Winjected-class-name).
 		 */
 		template<class UnitFrom, class UnitTo>
-		struct delayed_is_same_dimension : std::false_type
+		struct delayed_is_same_dimension_conversion_factor : std::false_type
```
johelegp: This is unused. If Clang 6 doesn't give warnings, as it documents (-Winjected-class-name), it can be removed.

PR #217, opened 2018-11-07
Add common type of time units and `std::chrono::duration`

johelegp:
Working towards correct `std::common_type` specializations. First, put everything in one place.

### DIFF
--- a/include/units/core.h
+++ b/include/units/core.h
@@ -2060,13 +2060,16 @@ namespace units
 						detail::is_non_truncated_convertible_unit<typename UnitFrom::conversion_factor,
 							typename UnitTo::conversion_factor>>>>;
 
+		template<class Ratio>
+		using time_conversion_factor = conversion_factor<Ratio, dimension::time>;
+
 		/**
 		 * @brief		SFINAE helper to test if a `conversion_factor` is of the time dimension.
 		 */
 		template<class ConversionFactor>
 		inline constexpr bool is_time_conversion_factor =
 			traits::is_same_dimension_conversion_factor_v<ConversionFactor,
-				conversion_factor<std::ratio<1>, dimension::time>>;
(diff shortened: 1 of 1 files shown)


johelegp:
> This is ready.

Actually, wait. This change and previous specializations also failed to meet this requirement: ["a program may specialize `common_­type<T1, T2>` for types `T1` and `T2` such that `is_­same_­v<T1, decay_­t<T1>>` and `is_­same_­v<T2, decay_­t<T2>>` are each `true`."](http://eel.is/c++draft/meta#trans.other-5.sentence-1). Partial specializations that use a template parameter as an argument in the _argument-list_ of the specialization, like this [`StrongUnit`](https://github.com/nholthaus/units/pull/217/files#diff-66d5febae528de60895295a0d564fbc3L394) or this [`T`](https://github.com/nholthaus/units/pull/217/files#diff-66d5febae528de60895295a0d564fbc3R374), could accept types that weren't the same if decayed. That would mean that for each strong unit of the same dimension, say `millimeters` through `kilometers`, you'd have to "manually" define partial specializations in pairs of all their combinations. And this has to expand to non-`meters` length units, like `feet`. This is unreasonable, and explodes the amount of symbols in the library even more. `std::common_type` has a really unfriendly interface for specializing in cases such as this.

What do you think about solving this be reverting making the units strong types?

johelegp:
Changed the scope of the PR to adding these `std::common_type` specializations between time units and `std::chrono::duration` specializations.

nholthaus:
@johelegp I'm a bit confused by the comment timeline between this and #219, but I think that we still want to pull this PR even if we revert strong units?

johelegp:
That's right.

nholthaus:
I made an attempt to resolve the conflicts but I wasn't confident in the results. Would you mind taking a stab at it?

PR #220, opened 2018-12-17
Setting Stefan-Botzman constant as a precalculated fixed value and a …

achary:
Even when #67 has been closed, here is a patch to improve how the library works with float as underlying unit type. Having single precision float as correctly supported type enables library to be usable for embedded systems, which is where I plan to use it.


achary:
Not sure why **continuous-integration/appveyor/branch** stucked here, and I see no link to go and check. Did that happened before?

PR #226, opened 2019-02-28
Make unarary ``operator+()`` and ``operator-()`` ``constexpr``

alexdewar:
These two functions can be safely marked ``constexpr``.

At present, if you try to write:
```
constexpr meter_t x = -1_m;
```
you get an error.

This PR fixes this.


johelegp:
Note to self: Check commits between `master..v3.x` that could be reasonably backported.

nholthaus (approved):

PR #227, opened 2019-04-09
Support negative exponent in power_of_unit

OpportunityLiu:


nholthaus:
This is great. We'll be pulling it once we cherry-pick it into the v3.x development.

PR #230, opened 2019-04-21
Add is_signed and is_integer to numeric_limits

jml1795:
Trying to use w/ the Eigen linear algebra package and these appear to be missing


nholthaus:
note to self: this is fixed differently in v3.x (inherit from std::numeric_limits<T>)

PR #251, opened 2020-10-05
revert: strong units

### LINKED ISSUE #219: Reverting #150, "strong" units
I believe merging #150 was an error.

For starters, its title is wrong. The units already were strong types, so there was nothing to "strengthen." The [README](https://github.com/nholthaus/units/blob/v3.x/README.md) is very clear on the advantages implied by this type safety, unlike using `double` or a type alias like `using meters = double;` for representing meters. I should have used some other term.

The purposes of the changes in #150 were to allow using CTAD and improved error messages. Out of this came the idea manifested in #167, which greatly took care of the latter purpose. And CTAD became the only purpose (personally) for #150.

Since then, we've discovered a few problems with the merged #150, as discussed in #200. In exchange for CTAD, we're paying `N * M` more in some ways. What used to be a single partial specialization of `std::common_type` for `unit`s, became `M` specializations for each of the `N` "strong" units.

Hopefully one day alias templates can support CTAD, as is being proposed in https://wg21.link/P1021. For now, I'd rather settle with being explicit about the underlying type. Naturally, I propose myself for reverting #150. And I am sorry for pushing for it without fully realizing what it entailed.

johelegp:
resolves #219.

Refs: 1634c7864d782e9ef960e73a45a376a4b728ca30

This ICEs on GCC 10+ due to [GCC 95486](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95486). So it may be incomplete.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ OPTION(BUILD_DOCS "Build the documentation" OFF)
 OPTION(DISABLE_IOSTREAM "Disables <iostream> (cout) support for embedded applications" OFF)
 
 SET(CMAKE_CXX_STANDARD_REQUIRED TRUE)
-SET(CMAKE_CXX_STANDARD 20)
+SET(CMAKE_CXX_STANDARD 17)
 
 # header-only library target. To use this project as a subdirectory,
 # add the following to your code:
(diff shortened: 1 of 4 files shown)


johelegp:
I'll specify the underlying type's as a template argument to make this C++17 again, like it was before the merge.

nholthaus:
thank you for this. It's a bummer that CTAD didn't work out. It seems to be a hard language feature to specify and implement.

nholthaus:
@johelegp the quality of your commit messages has been on my mind since yesterday... they're exquisite. Do you follow some system that I could read about?

johelegp:
I'm happy to hear that. It's https://www.conventionalcommits.org/.

Issue #209, opened 2018-11-05
remove named conversion factors


nholthaus:
Named conversion factors no longer appear in unit definitions, so I don't think we need them anymore if we change the macros around. It would save us a bunch of symbols, and remove an obsolete concept from the public interface.

I made a half-hearted attempt, but reverted when I ran in to trouble in the strong definitions.

johelegp:
Please, remind me what did you mean by "named" conversion factors.

nholthaus:
what used to be `meters` etc but is now called `meter_conversion_factor`

johelegp:
So you mean removing the strong conversion factors and leave only `conversion_factor`?

nholthaus:
I mean removing the entire `UNIT_ADD_UNIT_TAGS` macro

nholthaus:
so sometimes when you leave an issue open for two years, you don't really know what was going on and discover it wasn't a good idea. This is one of those. Not fully understanding the evolution of `strong` units didn't help.

This change regresses the error messages, but doesn't improve compile time. The readability gain in the definitions is more than offset by the readability loss to the user.

johelegp:
> The readability gain in the definitions is more than offset by the readability loss to the user.

The former can be retained. And I like it.

nholthaus:
How so, by refactoring the macros some more?

johelegp:
By using the unit's `conversion_factor` member alias.

johelegp:
Basically #256 without the bits that remove the strong conversion factors, thanks to this line:
```C++
 			abbreviation, ::units::decibel_scale, typename ::units::namespaceName::namePlural<UNIT_LIB_DEFAULT_TYPE>::conversion_factor) /** @} */ \
```
There are a few redundant `UNIT_LIB_DEFAULT_TYPE`s, though. The "diamond operator" `<>` does the job.

johelegp:
> so sometimes when you leave an issue open for two years, you don't really know what was going on and discover it wasn't a good idea. This is one of those. Not fully understanding the evolution of `strong` units didn't help.

I think I remember what happened. With strong units, `meters` or `meters<>` (actually `meters<double>`) could replace (strong) conversion factors. And error messages would've been even better. But strong units were reverted. I didn't put a reference to the comment where this idea spawned back then because I lost track of it.

nholthaus:
[...]

Issue #245, opened 2020-06-05
/permissive- definition


nairaner:
Defining `/permissive-` causes issues when compiling using Clion+CUDA. 

First of all `/permissive-` is compiler flag and not definition. PR #224 fixes it, but causes another issue. It forces standard conformance on library users, which can't always make it happen. I believe it should be hidden behind CMake option or just be removed from CMake and information about compilation time improvement available in README

https://github.com/nholthaus/units/blob/9f09abb32641fdf8dbb2c49ee81717d434624218/CMakeLists.txt#L69

nholthaus:
I just ran into this myself, I think it's a problem with CLion in general. I think you're right about imposing it on users (as the former only user of the library I took some liberties), so I'll leave it in some compile-time-improvement section of the readme.

nholthaus:
Resolved: `/permissive-` no longer appears in CMakeLists.txt; nothing is forced on consumers. Closing.

PR #256, opened 2020-10-06
V3.x simplify conversion factors

nholthaus:
This PR removes the symbols naming conversion factors for each unit container type.


CMakeLists.txt
```diff
@@ -1,86 +1,86 @@
...
-    DESTINATION share/units/cmake
-)
-
-install(DIRECTORY include/
-    DESTINATION include
-)
+cmake_minimum_required(VERSION 3.9 FATAL_ERROR)
+
+PROJECT(units LANGUAGES CXX)
+set(units_VERSION 3.0.0)
```
johelegp, line 4:
```suggestion
project(units VERSION 3.0.0 LANGUAGES CXX)
```

CMakeLists.txt
```diff
@@ -1,86 +1,86 @@
...
+
+# cmake options
+OPTION(BUILD_TESTS "Build unit tests" ON)
```
johelegp, line 7:
```suggestion
OPTION(BUILD_TESTS "Build unit tests" OFF)
```
This is more friendly to users. Developers, the CI and interested users should explicitly turn it on.

CMakeLists.txt
```diff
@@ -1,86 +1,86 @@
...
+OPTION(BUILD_DOCS "Build the documentation" OFF)
+OPTION(DISABLE_IOSTREAM "Disables <iostream> (cout) support for embedded applications" OFF)
```
johelegp, line 9: Consider prefixing `UNITS_` to the options to avoid clashes with other libraries.

CMakeLists.txt
```diff
@@ -1,86 +1,86 @@
...
+
+SET(CMAKE_CXX_STANDARD_REQUIRED TRUE)
```
johelegp, line 11: Similar to above, the library doesn't strictly require this. Also see #245.

CMakeLists.txt
```diff
@@ -1,86 +1,86 @@
...
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+	add_compile_options(-Wall -Wextra -pedantic)
+elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+	# enable parallel build
+	set( ENV{CL} /MP )
+#	target_compile_definitions(${PROJECT_NAME} INTERFACE "/permissive-")
+elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+	# Increase warning levels
+	add_compile_options(-Wall -Wextra -pedantic)
+endif()
```
johelegp, line 35: This, too, belong in the CI and developer's instructions. Then you can use `-Werror` without remorse.

CMakeLists.txt
```diff
@@ -1,86 +1,86 @@
...
+
+# Remove IOStream from the library (useful for embdedded development)
+if(DISABLE_IOSTREAM)
+	target_compile_definitions(${PROJECT_NAME} INTERFACE -DUNIT_LIB_DISABLE_IOSTREAM)
+endif(DISABLE_IOSTREAM)
+
+# unit tests
+if(BUILD_TESTS)
+	enable_testing()
```
johelegp, line 44:
```suggestion
include(CTest)
if(BUILD_TESTING AND BUILD_TESTS)
```
See https://cmake.org/cmake/help/latest/module/CTest.html.

.travis.yml
```diff
@@ -44,9 +44,9 @@ install:
 script:
   - mkdir build && cd build
   - cmake --version
-  - cmake ..
-  - cmake --build . --config Release
-  - cmake --build . --config Debug
+  - cmake _DUNITS_BUILD_TESTS=ON ..
```
johelegp, line 47:
```suggestion
  - cmake -DUNITS_BUILD_TESTS=ON ..
```

johelegp:
The change is really hard to see. Can you rebase it on `v3.x`?

nholthaus:
rebased to v3.x. coveralls isn't working (maybe hasn't been for a while?) but I don't want that to hold up the rename, and I think it would be better to solve it in another PR.

johelegp:
What rename? The units were renamed and this removes the conversion factors. I can tell you did that from the tests, but I can't see how you did it because the diff for `core.h` is messed up.

nholthaus:
Whitespace issue, I was hoping the rebase would help. I'll take care of that.

You can view the differ [without the whitespace change](https://github.com/nholthaus/units/pull/256/files?w=1)

include/units/core.h
```diff
@@ -1,3922 +1,3892 @@
...
+//	MACROS
+//------------------------------
+
+/**
+ * @def			UNIT_ADD_UNIT_DEFINITION(namespaceName,namePlural)
+ * @brief		Macro for generating the boiler-plate code for the unit type definition.
+ * @details		The macro generates the definition of the unit container types, e.g. `meter`
+ * @param		namespaceName namespace in which the new units will be encapsulated.
+ * @param		namePlural - plural version of the unit name, e.g. 'meters'
+ */
```
johelegp, line 130:
```suggestion
 * @param		__VA_ARGS__ the conversion factor of the unit, e.g. `meters<>`
 */
```

include/units/core.h
```diff
@@ -1,3922 +1,3892 @@
...
+ * @details		The macro generates the definition of the scaled unit templates as a strong type template alias,
+ *				e.g. `meters`
+ * @param		unitName unit name, e.g. 'meters'
+ * @param		scale the non linear scale template argument of the unit's base
+ * @param		definition - the variadic parameter is used for the definition of the unit
+ *				(e.g. `conversion_factor<std::ratio<1>, units::dimension::length>`)
+ * @note		a variadic template is used for the definition to allow templates with
+ *				commas to be easily expanded. All the variadic 'arguments' should together
+ *				comprise the unit definition.
+ */
```
johelegp, line 150: Like above.

include/units/core.h
```diff
@@ -1,3922 +1,3892 @@
...
+ * @param		namespaceName namespace in which the new units will be encapsulated. All literal values
+ *				are placed in the `units::literals` namespace.
+ * @param		namePlural - plural version of the unit name, e.g. 'meters'
+ * @param		abbreviation - abbreviated unit name, e.g. 'm'
+ * @param		definition - the variadic parameter is used for the definition of the unit
+ *				(e.g. `conversion_factor<std::ratio<1>, units::dimension::length>`)
+ * @note		a variadic template is used for the definition to allow templates with
+ *				commas to be easily expanded. All the variadic 'arguments' should together
+ *				comprise the unit definition.
+ */
```
johelegp, line 249: Like above.

johelegp:
How are the error messages after this? `meters` is actually `unit<...>`, so I suspect they'll revert the compiler to being a novelist.

nholthaus:
I need to check, when I started we still had strong units.

That said, conversion names didn't appear in the error messages as often a
you might think, hence why we used to see all the old SI dimension ratios.

johelegp:
> That said, conversion names didn't appear in the error messages as often a
you might think, hence why we used to see all the old SI dimension ratios.

Seems like I didn't implement the follow up to identify strong units within ephemeral units. I suspect it's not easy to do, anyways.

nholthaus:
The latest push keeps the readability improvements in the definition, while retaining the improved error messages. Clang is a bit inconsistent about when it uses named strong types, but overall still improved.

We weren't able to get rid of any symbols, but at least they've become a pure implementation detail. 

I think this is ready to go.

nholthaus:
resolves #209

PR #263, opened 2020-10-29
Fix some cmake option Variable and add an alternative cmake configuration

augustoicaro:
Hello!

I was configuring this library in my company Linux project today and found some inconsistencies in cmake variables. I took some time to be able to compile this library inside my project. I believe the problem was with some macros clash. After using macro definitions to use only the units that I was interested in, everything works.

I choose your library after a great presentation by Mateusz Pusz at CppCon2020, it is the best option for the project limitations. We are very excited to test it in production soon.

Congrats!


nholthaus:
awesome PR, thank you very much. I'll pull this once I roll everything into the dev branch

augustoicaro:
Thanks! Sorry if I didn't follow the PR branch template.

nholthaus (approved):

PR #265, opened 2020-11-03
pow() tweaks and a compile-time pow()

ts826848:
This has a little bit of cleaning up my previous pow() PR and also adds a pow() that uses a compile-time exponent.

### DIFF
--- a/include/units/core.h
+++ b/include/units/core.h
@@ -1707,7 +1707,7 @@ namespace units
 	namespace detail
 	{
 		template<typename T1, typename T2>
-		constexpr auto pow_acc(T1 acc, T1 x, T2 y)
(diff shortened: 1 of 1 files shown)


ts826848:
Fun fact: MSVC appears to produce an absolutely [*astounding* amount of assembly](https://godbolt.org/z/5rqe1W) for this particular compile-time pow(), the vast majority of which are seemingly pointless instantiations of `pow_acc`. In this case, `pow<1024>` caused MSVC to generate instantiations for `pow_acc<536870911, double>`, `pow_acc<9007199254740991,double>`, and `pow_acc<-1025,double>`.

Luckily, the actual function `f(double)` generated the expected assembly. The rest was certainly a surprise, though.

PR #310, opened 2023-03-03
cmake: Allow to use existing googletest package

krf:
There's no need to setup your own install in case there's one already.


nholthaus:
you're right, I love this.

PR #315, opened 2023-03-31
Fix compilation error when using units::sqrt with a float underlying type

Guillaume227:
(as opposed to double) - see #313 for context.


nholthaus (approved):

nholthaus:
note to self: couldn't merge due to conflict, added manually. Good PR.

PR #325, opened 2023-11-13
Avoid pascal macro conflict when working on mingw

### LINKED ISSUE #324: Compilation under MINGW
We use Units within our openpass. One of the goals is to provide openpass executable under windows without the need for MSVC. Therefore, it would be graeat if sources would be compatible to MINGW64 (MSYS2).

We are using the Units library (version 2.3.3), and we use gcc compiler to build openPASS. We have noticed that when working on Windows, "Pascal" macro name conflict would arise.

raghunandannm:
resolves #324

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -60,6 +60,11 @@
 #	endif // _MSC_VER < 1800
 #endif // _MSC_VER
 
+#if defined(__MINGW64__) || defined(__MINGW32__)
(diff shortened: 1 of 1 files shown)


JohelEGP:
To the opening comment, you should add "resolves #324".

reneparis:
Any updates?

Issue #328, opened 2024-01-19
Math functions not compatible with percent


StefanoD:
Please include the following information in your issue:
1. Which version of `units` you are using
Branch v3.x, Commit da6dd9176e8515323c75030d5e51ee19cf6c9afd

2. Which compiler exhibited the problem (including compiler version)
GCC 11.4.0

The math function don't work for dimensionless units like `percent`

```cpp
using units::concentration::percent;

percent<double> pct1 = 100_pct;
percent<double> pct2 = 70_pct;

percent<double> delta = fabs(pct1 - pct2); // Results into percent<double>(0.3)
percent<double> maxValue = fmax(pct1, pct2); // Results into percent<double>(1.0)
percent<double> minValue = fmin(pct1, pct2);  // Results into percent<double>(0.7)

// Following tests are failing:

EXPECT_EQ(delta, 30_pct);
EXPECT_EQ(maxValue, pct1);
EXPECT_EQ(minValue, pct2);
```

The problem lies in `core.h` where the said functions use `.value()` instead of `.raw()`

nholthaus:
`fabs`, `fmin`, and `fmax` operate on `.raw()` in the current line (`core.h`), so they work correctly on ratio-dimensionless units — e.g. `fabs(100_pct - 70_pct) == 30_pct`.

PR #329, opened 2024-01-19
Fix math functions for percent

StefanoD:
Fixes https://github.com/nholthaus/units/issues/328

Note: I tried to alter all math functions, but this resulted in some unit tests failing. There might be some rework needed from somebody with more knowledge.


nholthaus:
I think this may stem from a misunderstanding of how concentrations are intended to work. The following tests will pass in the library as it currently exists:

percent resultPct = sqrt(16.0_pct);
EXPECT_EQ(resultPct, 4.0_pct);
EXPECT_EQ(0.04, resultPct);
	
So the idea is that 4% when it's a dimensionless absolute value is 0.04.

chiphogg:
Shouldn't `sqrt(16.0_pct)` result in something like "4 perten" (made-up unit), which when converted to `percent` would result in `40.0_pct`, not `4.0_pct`?

StefanoD:
If you go to my linked ticket, I showed different calculations which failed. This is IMHO a regression which previously worked in an older version.
Anyway: We are using the Aurora units lib now.

PR #331, opened 2024-04-05
Some numeric_limits and numeric helper functions were missing

siposcsaba89:
Extended std::numeric_limits class with some functions needed to use it with yaml-cpp.

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -4854,9 +4854,53 @@ namespace std
 		{
 			return units::unit_t<Units, T, NonLinearScale>(std::numeric_limits<T>::lowest());
 		}
+		static constexpr units::unit_t<Units, T, NonLinearScale> epsilon()
+		{
+			return units::unit_t<Units, T, NonLinearScale>(std::numeric_limits<T>::epsilon());
+		}
+		static constexpr units::unit_t<Units, T, NonLinearScale> round_error()
+		{
+			return units::unit_t<Units, T, NonLinearScale>(std::numeric_limits<T>::round_error());
+		}
+		static constexpr units::unit_t<Units, T, NonLinearScale> infinity()
+		{
+			return units::unit_t<Units, T, NonLinearScale>(std::numeric_limits<T>::infinity());
+		}
+		static constexpr units::unit_t<Units, T, NonLinearScale> quiet_NaN()
+		{
+			return units::unit_t<Units, T, NonLinearScale>(std::numeric_limits<T>::quiet_NaN());
+		}
+		static constexpr units::unit_t<Units, T, NonLinearScale> signaling_NaN()
+		{
+			return units::unit_t<Units, T, NonLinearScale>(std::numeric_limits<T>::signaling_NaN());
+		}
+
 		static constexpr bool is_integer = std::numeric_limits<T>::is_integer;
 		static constexpr bool is_signed = std::numeric_limits<T>::is_signed;
+		static constexpr bool has_quiet_NaN = std::numeric_limits<T>::has_quiet_NaN;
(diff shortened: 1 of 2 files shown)


include/units.h
```diff
@@ -4854,9 +4854,53 @@ namespace std
 		{
 			return units::unit_t<Units, T, NonLinearScale>(std::numeric_limits<T>::lowest());
 		}
+		static constexpr units::unit_t<Units, T, NonLinearScale> epsilon()
+		{
+			return units::unit_t<Units, T, NonLinearScale>(std::numeric_limits<T>::epsilon());
+		}
```
BenFrantzDale, line 4860: Since this is epsilon near 1.0, I feel like it's not unit-invariant the same way `lowest()` is.

include/units.h
```diff
@@ -4854,9 +4854,53 @@ namespace std
...
+		return std::isinf(x());
+	}
+
+	template<class Units, typename T, template<typename> class NonLinearScale>
+	bool signbit(const units::unit_t<Units, T, NonLinearScale>& x)
+	{
+		return std::signbit(x());
+	}
+
 }
```
BenFrantzDale, line 4904: Should we provide https://en.cppreference.com/w/cpp/types/numeric_limits/is_specialized ?

include/units.h
```diff
@@ -4854,9 +4854,53 @@ namespace std
...
+		}
+		static constexpr units::unit_t<Units, T, NonLinearScale> quiet_NaN()
+		{
+			return units::unit_t<Units, T, NonLinearScale>(std::numeric_limits<T>::quiet_NaN());
+		}
+		static constexpr units::unit_t<Units, T, NonLinearScale> signaling_NaN()
+		{
+			return units::unit_t<Units, T, NonLinearScale>(std::numeric_limits<T>::signaling_NaN());
+		}
+
```
BenFrantzDale, line 4877: Consider adding `denorm_min` https://en.cppreference.com/w/cpp/types/numeric_limits/denorm_min

BenFrantzDale (commented):
I was hoping for this! I have to jump through hoops to get inf.

A few comments. Also, you should probably add tests in https://github.com/nholthaus/units/blob/master/unitTests/main.cpp

include/units.h
siposcsaba89: I am not sure, it depends on the underlying type, or I don't understand something,

include/units.h
siposcsaba89: Yes, you are right, added with the new commit.

include/units.h
siposcsaba89: Yes, that is missing too, included with the new commit

siposcsaba89:
> I was hoping for this! I have to jump through hoops to get inf.
> 
> A few comments. Also, you should probably add tests in https://github.com/nholthaus/units/blob/master/unitTests/main.cpp

Also added some unit tests for limits. Can you check them? Maybe you have some other tests in mind.

include/units.h
BenFrantzDale: My thought here is I can’t think of a case where you’d want epsilon that wasn’t better served by multiplying by `std::numeric_limits<double>::epsilon()`. Epsilon is the smallest increment you add to 1.0 to get a different number, but with units, 1.0 what? I can find the epsilon WRT `1.0_mm` by multiplying by double-epsilon, but I can’t think of a use case in which the number isn’t magic. Does that make sense?

BenFrantzDale: Like if I change `D` between meters and millimeters then `1.0_mm + std::numeric_limits<D>::epsilon()` would change truthiness, I think.

nholthaus:
great PR guys thanks!!

PR #336, opened 2024-08-21
Add #undef PI

codereptile:
Some other libraries define PI, which creates a conflict. This will implicitly resolve this problem. (It can be resolved explicitly, by adding this line of code before include, but that's extremely annoying)

### DIFF
--- a/include/units.h
+++ b/include/units.h
@@ -4162,6 +4162,7 @@ namespace units
 		 * @anchor constantContainers
 		 * @{
 		 */
+		#undef PI
 		using PI = unit<std::ratio<1>, dimensionless::scalar, std::ratio<1>>;
 
 		static constexpr const unit_t<PI>																											pi(1);											///< Ratio of a circle's circumference to its diameter.


codereptile:
Just wanted to say that this came from Unreal Engine, where it defines PI too

chiphogg:
I feel the pain here --- we recently hit the exact same problem in Au.  [This conversation thread](https://github.com/aurora-opensource/au/pull/247#discussion_r1647801986) may be relevant for considering the options for nholthaus/units.

Unfortunately, as that thread points out, even if this landed it would be a fragile solution, because it would depend on the order of includes.  As far as I can tell, if you're using a library that defines a macro for `PI` (such as Unreal Engine, or the Arduino core library), then there's just no good solution for defining the symbol `PI` _anywhere_ in your library.  We ended up [deprecating it](https://github.com/aurora-opensource/au/pull/250), and we'll remove it after it's been deprecated for a full minor release cycle.

It's certainly frustrating that people would define short-named macros like this in more modern and enlightened times, but if we want to support those libraries, I guess we're stuck with it!

codereptile:
@chiphogg Well, this wasn't much pain for me, a simple undef kind of solved it, but in the end I still switched to Au. By the way, I've started making a wrapper for UE with it, and so far it looks REALLY good. Especially since Unreal Engine uses centimeters as base unit for length, which is a pain in any physics formula. Being able to call methods and get results in any units turned out to be extremely useful. What do you think about this idea?)

chiphogg:
Wow, really glad to hear Au's working well for you!

> What do you think about this idea?

(Assuming "this idea" means making a wrapper for UE that uses Au)

If UE uses `cm` throughout for length, but has raw numeric types in its interfaces... oof.  In that case, yes, I think a set of "wrapper" APIs, which let users pass `Quantity` types (or receive them as return values), and then simply forward the underlying values to/from the UE APIs, would be the way to go --- I think it's what I'd do too.

codereptile:
@chiphogg Yea "oof" kind of sums it up. I was wondering if we could advertise that as one of the examples for Au usage.

chiphogg:
Yeah, that sounds like a really compelling use case!

PR #337, opened 2024-08-28
Updated CMake policy version to resolve deprecation warning.

joe-barnier:
From <https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html>:

> Changed in version 3.27: Compatibility with versions of CMake older than 3.5 is deprecated. Calls to cmake_minimum_required(VERSION) or cmake_policy(VERSION) that do not specify at least 3.5 as their policy version (optionally via ...<max>) will produce a deprecation warning in CMake 3.27 and above.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.2...3.5 FATAL_ERROR)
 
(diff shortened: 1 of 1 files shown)


joe-barnier:
No idea why the continuous-integration/appveyor/pr fails (and it also seems to be failing for other recent PR's).

Is that something someone else is looking into? Is it something I need to look into if I want this PR merged?

nholthaus (approved):

PR #338, opened 2024-09-13
Added namePlural to UNIT_ADD_NAME

raulcalvo:
I found no way of obtaining the namePlural in the same way it is done with name, so I added it.


nholthaus:
note to self: approved, need to test locally prior to merge

PR #356, opened 2026-07-16
Fix std extensions still using `operator()`

carlfriess:
`std::isnan()`, `std::isinf()` and `std:: signbit()` were still using the legacy `operator()` and `std::isfinite()` was missing. I also fixed the unit tests so that they can run on macOS.

Thanks for the awesome library!


nholthaus:
closing: this got merged as part of 3.4.0. Thank you for the great contribution!

PR #358, opened 2026-08-14
v3.4.0: readable diagnostics (class-based named units) + #357 fix + natvis + error-msg CI harness

### LINKED ISSUE #357: (v3) Creating an expression that simplifies to a certain dimension before including the dimension's header causes explicit specialization errors
MRE:
```
#include <units/velocity.h>
#include <units/length.h>
auto x = 0_mps / 0_m;
#include <units/frequency.h>
```

This is particularly problematic when going between linear and angular velocities, since you'll briefly cancel out the distance unit before applying a new distance unit, so one of the expressions will result in Hz, effectively, which then conflicts with the frequency definitions.

nholthaus:
## Summary

A **readability + robustness** release for the v3.x line — focused on what you see in the compiler and the
debugger. **No public API changes**: `meters<>` / `meters<double>` and every unit spell exactly as before.

### 1. Diagnostics name the unit, not the template soup
Named units are now classes deriving from `units::unit<...>` (instead of alias templates), so a diagnostic
reads `units::length::meters<double>` → `units::frequency::hertz<double>` instead of
`unit<conversion_factor<std::ratio<1>, dimension_t<...>>, double, linear_scale>` — for **both operands and
computed results** (`length*length` → `square_meters<double>`, etc.). Rendering only; representation/values/
API unchanged.

### 2. Fix #357 — include order no longer breaks the build
Forming an expression that reduces to a not-yet-included dimension (`velocity/length` → `1/time`) failed with
*"explicit specialization after instantiation."* The strong-type lookup is now an ADL customization point, so
a later-included header just adds a better overload — include order can't decide compilability.

### 3. Visual Studio natvis
`natvis/units.natvis` shows a unit as `3 m` in Watch/Locals/DataTips; attached to the target + installed
(MSVC-only, no-op elsewhere).

### 4. `std::` math extensions (upstream PR #356 by @carlfriess, merged into this branch)
`std::isnan/isinf/signbit` switched off the removed `operator()` to `raw()`; adds `std::isfinite`; macOS
locale branch for the locale test.

## Testing
- `unitLibTest` **238/238** green, incl. new `namedUnitReportedTypeIsPreserved` + `unitsAsContainerKeys`.
- New **error-message harness** (`test/errorMessages/`, 29 cases) grading readable-named-type presence and
  no-`conversion_factor`-soup — compiler-portable (g++/clang/MSVC) and wired into all three CI workflows.

## Under the hood
Trait/`std` specializations, arithmetic operators, `to_string`/`operator<<`, `name()`/`abbreviation()` all
see through / preserve the named class; `numeric_limits<Named>` returns the named type; `common_type
<dimensioned-named, scalar>` is SFINAE-empty (parity with the plain form); pure `dimensionless` stays a plain
alias (interchangeable with int/double).

Version bumped to **3.4.0**.

### DIFF
--- a/.github/workflows/clang-19.yaml
+++ b/.github/workflows/clang-19.yaml
@@ -45,3 +45,7 @@ jobs:
       - name: Run Tests
         working-directory: build
         run: ctest --output-on-failure
+
+      - name: Error-message tests
+        run: |
+          python3 test/errorMessages/run.py --cc clang++-19 --std c++23 --include include
--- a/.github/workflows/gcc-13.yaml
+++ b/.github/workflows/gcc-13.yaml
@@ -44,3 +44,7 @@ jobs:
       - name: Run Tests
         working-directory: build
         run: ctest --output-on-failure
+
+      - name: Error-message tests
+        run: |
+          python3 test/errorMessages/run.py --cc g++-13 --std c++23 --include include
--- a/.github/workflows/msvc-2022.yaml
+++ b/.github/workflows/msvc-2022.yaml
@@ -15,14 +15,27 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v4
 
+      # Set up the MSVC command-line environment FIRST (puts cl.exe on PATH). The "Visual Studio 17 2022"
+      # generator's own VS discovery has been unreliable on recent windows-latest images ("could not find any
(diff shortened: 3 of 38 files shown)


chiphogg:
Readable compiler errors with concise typenames?  That's a huge upgrade!  Congrats on this improvement --- it's awesome to see. :rocket:

PR #362, opened 2026-08-14
docs: spelling a quantity type (meters / meters<> / meters<T>) and when to use auto

nholthaus:
Adds an early README subsection covering two things new users ask immediately:

- **Spelling the type** — `meters<T>` (explicit representation) and `meters<>` (default) are valid as variables, parameters, return types, and members; bare `meters` is a CTAD placeholder valid only where an initializer is present to deduce from (so not as a parameter, return type, or member). Includes the ill-formed cases with the reason.
- **`auto` vs. an explicit type** — use `auto` when the right-hand side states the unit; name the result type explicitly when the compiler should confirm the dimensional analysis (a mismatch becomes a compile error instead of an accepted `auto` deduction).

Every snippet compiles. Also removes a "surprising" characterization from the int/double CTAD note in the CTAD/ADL guide (states the rule as fact).

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,40 @@ meters d{5.0};          // braced construction
 > arithmetic, so `1_m / 2_m` is `0`, whereas `1.0_m / 2.0_m` is `0.5`. Use a decimal point (or write
 > `meters<double>`) when you want fractional results.
 
+**Spelling the type: `meters`, `meters<>`, `meters<T>`.** Three ways to name the type:
+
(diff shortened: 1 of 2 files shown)


nholthaus:
Folded into #364 (the single v3.4.2 PR) as commit b2599c9 — the meters/meters<>/meters<T> spelling guidance and the auto-vs-explicit-type section now ride on that branch. Reviewing everything as one PR.

PR #377, opened 2026-08-16
Add opt-in absolute<>/delta<> point/delta wrappers

nholthaus:
## Opt-in `absolute<>` / `delta<>` point/delta wrappers

Two thin, opt-in class templates that make the affine **point vs delta** distinction explicit in the type — the inverse of a point/delta-centric library. Plain unit types are **completely unchanged**; you reach for the wrappers only where you want the compiler to enforce the distinction (temperatures, epochs vs durations, absolute vs gauge pressure, positions vs displacements).

Stacked on #376 (uses its affine-correctness work + `is_affine_unit_v`). Merge #376 first.

### The types
- **`absolute<U>`** — a POINT (carries the unit's datum). `.to<V>()` is a point conversion, offset applied (`0 °C → 273.15 K`).
- **`delta<U>`** — an AMOUNT (offset-free). `.to<V>()` is scale-only (`10 °C delta → 18 °F delta`, never `50 °F`).

They wrap **any** unit; for a non-affine unit the datum is zero, so `absolute` and `delta` coincide.

### Type algebra (compile-time enforced)
| Expression | Result |
|---|---|
| `absolute − absolute` | `delta` (datum cancels) |
| `absolute ± delta` | `absolute` (move the point) |
| `delta ± delta` | `delta` |
| `delta * / scalar`, `-delta` | `delta` |
| `absolute + absolute` | **ill-formed** (sum of two points is meaningless) |
| wrapper `±`/`==` plain unit | **ill-formed** (a plain unit is neither point nor delta here) |

Constructors are `explicit` (no implicit unit↔wrapper conversion). Wrappers are **trivially copyable and `sizeof == sizeof(U)`** — zero overhead.

### Correctness detail
Per-operation unit choice is deliberate: a **point stays in its own unit** (reconciling two affine frames to a "common" unit mangles the datum), while offset-free **delta arithmetic reconciles to the common/finer unit** so a mixed integer delta never narrows.

### Testing — an adversarial wart hunt
Three suites authored by fanning adversarial reviewers across the factorial, per "test the hell out of it, find warts preemptively":
- **AffineWrapper** — temperature point/delta algebra, scale-only delta conversion, no-datum-leak, constexpr.
- **GeneralWrapper** — non-affine + the pi-carrying angular units (verified the delta scale-only path preserves the pi exponent), integer underlying, cross-dimension.
- **WrapperSafety** — 70+ `static_assert`s: trivial-copyability, explicit ctors, plain/wrapper and wrong-dimension and point+point rejection via SFINAE, cmath/format/hash non-support, constexpr.
- A compile-fail `errorMessages` case for `absolute + absolute`.

The hunt found and fixed three real warts (int-delta×fraction truncation → now promotes; lossy cross-unit reconciliation → finer-common-unit; affine-point datum mangling → point stays in its own unit).

**Full suite green: 393/393 on g++ (C++23) and MSVC cl.exe** (verified via WSL interop). Documented in a new README section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,36 @@ Comparisons (`==`, `<`, …) and conversions between affine units are exact and 
 units (lengths, masses, everything without an offset) are unaffected — they add, subtract, and combine with
 no special cases.
 
+#### Making the point/delta distinction explicit: `absolute<>` and `delta<>`
+
+The rules above are enforced at run time by convention. When you want the compiler to enforce them, wrap a
+quantity in **`absolute<U>`** (a point on a scale — carries the datum) or **`delta<U>`** (an amount —
+offset-free). These are opt-in: plain unit types are unchanged, and you reach for the wrappers only where
+the safety matters (temperatures, epochs vs durations, absolute vs gauge pressure, positions vs
+displacements). They wrap *any* unit — for a non-affine unit the datum is zero, so `absolute` and `delta`
+coincide.
+
+```cpp
+using namespace units::temperature;
+
+absolute<celsius<double>> room(celsius<double>(20.0));
+delta<celsius<double>>    step(celsius<double>(5.0));
+
+auto d      = room - absolute<kelvin<double>>(kelvin<double>(273.15)); // delta<>  (20 K difference)
+auto warmer = room + step;                                            // absolute<> (25 degC)
+room       += step;                                                    // move the point in place
+
+// enforced at compile time:
+// auto bad = room + room;   // ill-formed — the sum of two points is meaningless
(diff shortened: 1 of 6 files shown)


nholthaus:
Superseded by the reworked affine wrappers landing in 3.6: the absolute/delta types are being reintroduced under the units::kind inline namespace, with the wart-hunt fixes (LHS-unit result reconciliation, delta compound assignment, is_absolute/is_delta traits and concepts, printing, min/max/abs) and built on top of the mixed-unit reconciliation changes (#381). This original single-commit version predates all of that. Closing in favor of the new PR; the branch can be deleted.

Issue #378, opened 2026-08-16
Subset-based headers in 3.x create strong risk of ODR violations and UB


chiphogg:
Two awesome new features in 3.x --- strong named types for units, and split-out headers --- combine to cause a bug in the current implementation.

The [subset headers instructions](https://github.com/nholthaus/units/blob/main/docs/how-to/subset-headers-compile-time.md) say that you must include the header for the _result_ types (e.g., velocity, if you divide length by time), which is what caught my eye.  This policy strikes me as untenable, because it requires users to notice all the ways that their quantities can combine, including intermediate results that they may not even notice.  So we should assume it will be routinely violated in practice, and ask: what if it is?

I believe the reason the policy exists is so that we can map the results of computations to the "canonical" unit: for example, mass times acceleration maps to Newtons, not kg m/s^2.  In practice, if anyone forgets to include the file, some locations will get Newtons, and some will get kg m/s^2.  This is a straightforward One Definition Rule (ODR) violation, which makes the entire program's behavior undefined.

I think the fix is straightforward.  Just stop making a registry of canonical units.  Let all unit operations produce the natural composition of their input types.  A quantity of kg m/s^2 is still _assignable_ to a quantity of Newtons, so people get Newtons in a natural way, by _asking_ for it.  As a matter of comparison, mp-units used to have this kind of canonical unit registry, but [removed it](https://mpusz.github.io/mp-units/latest/blog/2023/09/24/whats-new-in-mp-units-20/#the-downcasting-facility) in part because of ODR violations; Au never had one.  In my experience, this doesn't really hurt compiler error readability, because _stored_ types are unaffected; only _intermediate computations_ change, and even those are far more readable with your new strong-type units than with the old alias approach.

Here's an example to show the UB, which I ran under clang 20.1.4:

```cpp
// shared.h
#pragma once
#include <units/length.h>
#include <units/time.h>
#include <string>
inline std::string describe(double m, double s) {
    return units::to_string(units::meters<double>(m) / units::seconds<double>(s));
}

// a.cpp
#include <units/velocity.h>   // this TU knows the named type
#include "shared.h"
std::string from_a() { return describe(5.0, 2.0); }

// b.cpp — same, minus the velocity.h include
#include "shared.h"
std::string from_b() { return describe(5.0, 2.0); }

// main.cpp
#include <iostream>
#include <string>
std::string from_a();
std::string from_b();
int main() { std::cout << "a: " << from_a() << "\nb: " << from_b() << "\n"; }
[...]

PR #381, opened 2026-08-17
Reconcile mixed-unit arithmetic to a real named unit; recover composed and offset-free names

nholthaus:
## Summary

Mixed-unit arithmetic and reconciliation now land on the unit a reader expects, instead of an anonymous conversion-factor sub-unit.

- **`+` / `-` yield the LEFT operand's unit** — the caller controls the result unit by operand order and the value reads in the unit they wrote: `10 kg - 5 lb` is `7.73 kg`, `meters - feet` is meters, `feet - meters` is feet. The underlying widens to the common lossless unit only when the left operand is integral and cannot hold the right without truncation, so integer comparisons and modulo keep their exact reconciliation.
- **Affine subtraction reconciles to the left operand's scale** — `celsius(100) - fahrenheit(32)` reads as `100` celsius-degrees, not a value in an anonymous sub-unit of the two scales.
- **`std::common_type` recovers friendly names it previously missed** — a reconciled result equivalent to a named operand registered in *composed* form (centimeters = `centi<meters>`, minutes = 60 seconds) recovers that name (`meters`+`centimeters` → centimeters, `hours`+`minutes` → minutes). Recovery is gated to genuinely unnamed factors, so a factor that already carries a name (torque's `newton_meters`, energy's `joules`) is never renamed to another kind sharing its dimension and ratio, and `common_type` stays symmetric.
- **Affine-dimension reconciliation drops an arbitrary datum** — reconciling degrees Celsius with kelvin yields kelvin (the offset-free base a reader expects); units that share the offset keep it. The magnitude and pi ratios keep their greatest common measure, so integer exactness is preserved.

## Tests

- Full suite green (387) on g++; both g++ and clang exercised across the edge units (angle/pi, concentration, temperature/affine, time).
- New `arithmeticResultIsLeftOperandUnit` and `commonTypeRecoversComposedName` cover the left-operand rule across floating/integer and commensurable/incommensurable operands, the composed-name recovery, the affine offset drop, cross-kind symmetry, and preserved integer exactness (`km<int> == m<int>(1000)`, `m<int> != ft<int>(3)`).

## Compatibility

Behavior change (minor version): a mixed-unit `+`/`-` result type is now the left operand's unit rather than an anonymous common sub-unit; values are unchanged and remain losslessly assignable. Comparisons/`%`/`fmod`/`min`/`max` are unchanged (still reconcile via the symmetric common type).

### DIFF
--- a/include/units/core.h
+++ b/include/units/core.h
@@ -2425,6 +2425,20 @@ namespace units
 				std::conjunction<std::negation<std::is_floating_point<typename UnitFrom::underlying_type>>,
 					is_non_truncated_convertible_unit<typename UnitFrom::conversion_factor, typename UnitTo::conversion_factor>>>>;
 
+		/// True when both units have a floating-point underlying type.
+		template<class L, class R>
+		inline constexpr bool both_floating_v = std::is_floating_point_v<typename traits::unit_traits<L>::underlying_type> &&
+												std::is_floating_point_v<typename traits::unit_traits<R>::underlying_type>;
+
+		/// The result unit of a same-dimension `+`/`-` of `L` and `R`. It is the LEFT operand's unit — so the caller
+		/// controls the result unit by operand order and reads the value in the unit they wrote — whenever that is
+		/// lossless: either the right operand converts into the left as-is, or both operands are floating point (a
+		/// floating result in either unit loses nothing). When neither holds — an integral left operand that cannot
+		/// hold the right without truncation — it falls back to the symmetric common (finest, lossless) unit, the
+		/// same exact reconciliation integer comparisons use.
+		template<class L, class R>
+		using lhs_result_unit_t = std::conditional_t<is_losslessly_convertible_unit<R, L> || both_floating_v<L, R>, L, std::common_type_t<L, R>>;
+
 		// The underlying type a NAMED unit's from-unit deduction guide should produce when constructed from `Source`:
 		// the source's own underlying when losslessly convertible into the target (StrongCf, Scale), else its
 		// floating-point promotion (so e.g. radians(degrees{1}) deduces radians<double>). A SFINAE-friendly class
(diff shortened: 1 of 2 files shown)


chiphogg:
Just some food for thought: I think most users would expect addition to be commutative.  Taking the LHS's unit feels a little arbitrary.

The advantage I can see is that it produces results in friendly named units, not anonymous common units, more often.  It's also true that you get the correct _quantity_ result, regardless of the units it happens to be expressed in, so it's not a "bug" _per se_, just a light best practices violation.

I assume the main motivating disadvantages of the common unit approach are compiler errors, and unit labels.  On compiler errors, #378 has some future exploration ideas that could make those better.  On unit labels, there are neat things you can do with automatically expressing the common unit in terms of each of its distinct input units; see [aurora-opensource/au#105](https://github.com/aurora-opensource/au/issues/105) and [aurora-opensource/au#411](https://github.com/aurora-opensource/au/issues/411) for more details.

Hope these resources and feedback are helpful!

nholthaus:
It doesn't really matter because all the conversions are implicit and
resolved at compile time. If you specify the storage type you'll get
whatever type you want. I think it's less surprising for a user for an auto
result to deduce ONE of the source units in the computation rather than
some ephemeral unit, and the lhs convention gives the user control over
which unit that is and it only applies in an auto deduction scenario anyway
which itself is not a best practice.

The addition IS commutative, you get the same magnitude and dimension
regardless of the order of operands. 12 in == 1 ft regardless of which unit
is actually providing the storage and it's a considered design decision not
to draw a distinction between two equivalent values.

The raw value in this case is still well specified, and I encourage users
to avoid invoking things like .value() or arithmetic casts at all, unless
they have some physics function whose definition doesn't conform to strict
dimensional correctness (small angle assumptions, etc)

Long story short: it's a feature, not a bug.

PR #387, opened 2026-08-17
fix: unit conversions no longer overflow the intermediate computation

nholthaus:
A mul-then-divide conversion computed `value * num / den` with the intermediate `value * num` capped at the widest standard integer (or plain `double`). For a large magnitude that product could overflow *before* the divide recovered a value that fits the target — silently returning a wrapped integer or infinity even when the true result was representable.

## Fix
- **Integer intermediate:** carry `value * num / den` in a double-width integer — `__int128` where the compiler provides it, otherwise a portable 128-bit multiply-then-divide — so the product cannot overflow.
- **Floating-point intermediate:** keep the correctly-rounded `value * num / den` order for every normal magnitude; only when `value * num` would overflow to infinity fall back to divide-first (`value / den * num`), which returns the representable result instead of `inf`.

Normal-magnitude conversions are unchanged and stay correctly rounded (verified ≤ 0.5 ULP). The lossy divide-first path is taken only at the overflow threshold, so precision is never traded away when it isn't necessary.

## Tests
- Integer: `5e16 ft → m` = `1.524e16` (previously overflowed `int64`); small/negative/zero exact.
- Floating point: correctly-rounded on a spread of magnitudes, round-trip bit-stable; `1e306 ft → m` = `3.048e305` finite (previously `inf`).
- 390/390 on g++, clang-19, and MSVC (MSVC exercises the portable 128-bit fallback).

### DIFF
--- a/include/units/core.h
+++ b/include/units/core.h
@@ -2193,6 +2193,74 @@ namespace units
 
 	inline constexpr linearized_value_t linearized_value{};
 
+	/** @cond */ // DOXYGEN IGNORE
+	namespace detail
+	{
+		/// The widest integer available for a conversion's intermediate. A mul-then-divide conversion
+		/// (`value * num / den`) must not overflow the intermediate `value * num` before the divide recovers a
+		/// value that fits the target; a double-width intermediate holds the product. `__int128` is used where the
+		/// compiler provides it; otherwise the intermediate falls back to `std::intmax_t` (the widest standard
+		/// integer) and a manual 128-bit mul-divide guards the product.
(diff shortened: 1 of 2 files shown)


nholthaus:
Folded into #386 (the affine/kind + core-hardening bundle) so the affine wrappers receive the same overflow fix; no separate PR needed.

chiphogg:
Quick question: for floating point intermediate, why not compute `num / den` as a single floating point number, and just multiply by that?

You might want to make exceptions when `num` is exactly 1, because then you could just divide by `den`, but otherwise it seems like single-factor would be best.

PR #424, opened 2026-08-28
fix: complete the point-versus-change model for affine and decibel scales, with a named remedy for every rejection (#402, #409)

### LINKED ISSUE #402: (from Codex) Convert mixed affine deltas without applying their datum
When the right operand uses another affine scale, the non-deduced parameter implicitly converts it as an absolute point before this addition reads `rhs.raw()`. For example, `celsius<>{20} += fahrenheit<>{9}` produces approximately `7.22 °C`, because `9 °F` first becomes `-12.78 °C`, even though this operator documents the operand as a relative delta and should add 5 Celsius degrees to produce `25 °C`. Convert the right operand's scale factor as a delta while omitting its translation.

### LINKED ISSUE #409: (from Codex) Return an offset-free result from fdim
When either operand is affine, the positive difference is constructed as another affine point, so its datum is reapplied when the result is converted. For example, `fdim(celsius<>{100}, fahrenheit<>{32})` should represent a 100-degree Celsius/Kelvin delta, but converting the returned value to Celsius yields `-173.15` because the result represents an absolute 100 K point. As with the affine subtraction operator, return an offset-stripped delta unit.

nholthaus:
Makes the **point-versus-change** distinction complete and self-explaining across both scales that carry one — affine temperature and decibel. Every ordinary operation works; every operation that cannot work says why and what to write instead. **No breaking change** — `kelvin + kelvin` and all existing arithmetic keep working.

### The rule

A **reading** carries a datum or reference (`celsius`, `fahrenheit`, `dBW`, `dBm`); an **amount** does not (a difference, a scaled reading, an offset-free scale like `kelvin`, or a dimensionless dB gain). Which one you have is in the type, so the algebra follows from it.

**Affine temperature**

| You write | Result | |
|---|---|---|
| `c += celsius(5)` / `+= fahrenheit(9)` / `+= kelvin(5)` | reading, moved | **#402** — the rhs is an amount; its datum is never applied |
| `t1 - t2` | **amount**, in the left operand's degrees | the datums cancel (**#409**) |
| `c + amount`, `amount + c` | reading, moved | **new** — the by-value counterpart of `+=`, previously had no viable overload |
| `c * 2.0`, `2.0 * c`, `c / 2.0` | **amount** in its own degrees | **new** — scaling a reading has no meaning; scaling the amount does. Commutative. |
| `kelvin + kelvin`, `kelvin * 2.0` | kelvin | an offset-free scale is a magnitude — **unchanged** |
| `c1 + c2` | disabled | subtract for an amount, or move one with `+=` |
| `c *= 2.0`, `c += 5.0`, `c += meters(1)` | disabled | the in-place form cannot hold an amount; a number states no unit; different dimensions |

**Decibel** — the by-value operators already implemented this split; the in-place ones did not.

| You write | Result | |
|---|---|---|
| `level += gain`, `level -= gain` | level, moved | **new** — `level + gain` worked by value, but the compound form was rejected as a dimension mistake |
| `gain + gain`, `level - level`, `level ± gain`, `++level` | as documented | unchanged |
| `level *= 2.0`, `level /= 2.0`, `gain *= 2.0` | **disabled** | **was a silent wrong answer** — see below |
| `gain += 2.25` | disabled, at the call site | **was an error inside the library** |
| `level + level`, `level * 2.0` | disabled | unchanged |

### The bug worth reading about

`dBW(12.5) *= 2.0` compiled and returned **13.98 dBW** — neither a naive `25 dBW` nor a linear doubling's `+3.01 dB`. The body took the value **through** the scale (`raw()` is the dB number) and wrote it back **past** the scale (the `linearized_value` tag stores unlinearized). Those are the same operation only for a linear scale: 12.5 doubled to 25, then stored as the *linear* value. The by-value `operator*` already required `has_linear_scale_v`, so `dBW * 2.0` was correctly rejected while its compound form silently corrupted; `operator++` next door is right, rebuilding through the ordinary constructor. Now the in-place forms carry the same constraint as the by-value ones.

### Every rejection names its remedy

```
units: cannot add two affine points; subtract them for a units::delta<...>, or move one with +=.
units: cannot scale an affine point; scale a units::delta<...> instead.
units: cannot scale a decibel value; scale the linear quantity instead.
units: cannot add a bare number to a decibel value; add a decibels(...) gain.
units: cannot add a bare number to a quantity; add a quantity of the same dimension (e.g. meters(5)).
units: cannot multiply in place by a quantity; the product has a different dimension. Use `auto p = a * b;`.
```

`absolute<>` / `delta<>` remain **strictly opt-in**, for code that wants reading-vs-change enforced in the type (including on an offset-free scale, the one distinction a bare type cannot carry). Nothing requires them.

### Why the rules key on the datum, not the dimension

An earlier iteration keyed them on the *dimension*. That cannot work: a temperature **change is the same dimension** as a reading, so the point rules caught changes too and broke scaling-a-change and moving-a-reading-by-a-change. And nothing intrinsic separates a reading on an offset-free scale (`kelvin`) from a change — both are temperature with no datum — so a bare type cannot carry that distinction at all. Keying on the datum (which *is* in the type) makes the model expressible and keeps `kelvin` a magnitude, so no existing code breaks.

### Tests

- A temperature **operation matrix**: readings, offset-free scales, and changes against construction, conversion, comparison, compound move, difference, scaling, and by-value move — values and result kinds pinned.
- A **decibel matrix**: level/gain algebra, the dBm/dBW 30 dB reference offset, and the independence of the decibel *numerical scale* from the affine *datum*.
- The **integer-backed affine path**, where a cross-scale move promotes to floating point for the ratio and casts back (a 9 °F change is a 5 °C change, which must not truncate to zero).
- A **regression block** proving lengths, named angles, dimensionless and percent are untouched.
- **13 `absolute<>`/`delta<>` wrapper tests**.
- **13 errorMessages harness cases** requiring each disabled operation to fail **and** grading its message.

Also fixes **every warning in the test suite**: 40 `-Wshadow` sites were locals named after unit abbreviations, which are public API constants (`units::m`, `units::b` — `UNIT_ADD_CONSTANT`), now given descriptive names; the lone `-Wunused-function` was an overload existing only to make an overload *set* real, so rather than silence it the test now asserts it is a viable candidate — without which its companion assertion passed vacuously.

Adds **`scripts/ci_local_msvc.cmd`**: the MSVC leg had no local mirror, so local could pass where CI would not — and MSVC differs on exactly what this change touches.

Verified on **gcc-13** (Release + UBSan), **clang-19**, and **MSVC 2022**: 521 tests, zero warnings, diagnostic harness **90/90** on both gcc and cl.

Closes #402
Closes #409


nholthaus:
This has changed a bit since the initial conception:

### No Longer Compiles — all decibel

  | expression | 3.6.1 gave | why wrong |
  |---|---|---|
  | `dBW *= 2.0` / `/= 2.0` | 13.98 / 7.96 dBW | read thru scale, written past it |
  | `decibels *= 2.0` | 8.13 dB | same |
  | `quantity *= decibels(3.0)` | `meters(3)` → 9 m | multiplied by the dB figure |
  | `log10(decibels(3.25))` | 0.512 | ratio is 2.113, log 0.325 |
  | `atan(decibels(3.25))` | 1.2723 rad | +9 more functions |
  | `tanh(decibels(3.25))` | 0.99700 | same |
  | `delta<dBW>` / `absolute<dBW>` | wrapped | dimension already splits level/gain |

  ### Value/type changes, no compile error

  | | 3.6.1 | branch |
  |---|---|---|
  | `celsius(20) += fahrenheit(9)` | 7.22 °C | **25 °C** (#402) |
  | `celsius(20) += kelvin(5)` | −248.15 °C | **25 °C** |
  | `round<celsius<int>>(fahrenheit<int>(54))` | 30 °C | **12 °C** |
  | `hash(celsius(0))==hash(kelvin(273.15))` | false | **true** |
  | `numeric_limits<dBW>::max()` | `inf` | 3082.5 |
  | `numeric_limits<dBW>::epsilon()` | `0` | 9.6e-16 |
  | `decibels g; g = 3.25;` | 5.12 dB | **3.25 dB** |
  | `fdim(celsius(30),celsius(10))` | raw 20, **type=reading** → 293.15 K | raw 20, **type=amount** → 20 K |

  ### Newly works

  | | result |
  |---|---|
  | `celsius + amount`, `amount + celsius`, `celsius + celsius` | 30 °C |
  | `midpoint(c,c2)` / `lerp(c,c2,0.25)` | 15 / 17.5 °C |
  | `dBW += decibels(3.25)` / `-=` | 15.75 / 9.25 dBW |

  ### Unchanged from `main` — affine scaling & magnitude

  | | | | |
  |---|---|---|---|
  | `c * 2.0` → 40 | `c *= 2.0` → 40 | `c * percent(50)` → 10 | `abs(celsius(-5.25))` → 5.25 |
  | `c / 2.0` → 10 | `-c` → −20 | `sqrt(c)` → 4.4721 | `hypot(c,c2)` → 22.361 |
  | `c / c2` → 2 | `fmod(c,c2)` → 0 | `++c` → 21 | `sqrt(c*c)` → 20 |

  ### New public API

  | name | what |
  |---|---|
  | `traits::has_arbitrary_origin_v<U>` | affine reading **or** dB level; public because `requires` can't see body-fired refusals |
  | `traits::is_decibel_level_v<U>` | dimensioned=level, dimensionless=gain |
  | `units::lerp` / `units::midpoint` | `a+(b-a)*t`; datum-independent iff weights total 1 |

  ### Design Decisions

  | # | decision | status |
  |---|---|---|
  | 1 | `a - b` ≠ `a -= b` for offset-free rhs | user can disambiguate with `<absolute>` |
  | 2 | scaled reading valid only in its own scale (`20 °C×2`=40 °C vs `293.15 K×2`=313.15 °C) | documented, not refused |
  | 3 | by-value `dBW * 2.0` | refused |

Issue #440, opened 2026-09-11
squared/cubed/square_root produce an unnamed conversion factor, so sqrt(squared<U>) is not U


nholthaus:
`square_root<squared<U>>` is not the same type as `U`, for any unit:

```cpp
using rootMeters = unit<square_root<squared<units::length::meters_>>, double>;
static_assert(std::is_same_v<meters<double>, rootMeters>);   // fails
static_assert(std::is_convertible_v<rootMeters, meters<double>>);   // holds
```

The value and the conversion ratio round-trip exactly, and the result is implicitly convertible to `U` and compares equal to it, so `meters<double> m = sqrt(metre * metre);` works. Only the spelling differs. The same is true of `squared<meters_>`, which is not `square_meters_`.

### Why

`squared_impl`, `cubed_impl` and `sqrt_impl` build a bare `conversion_factor<...>`. The library recovers the *named* type from a structural one with `traits::strong_t`, an ADL customization point registered per unit by `UNIT_ADD_STRONG_CONVERSION_FACTOR`; `common_type`, `dimensionless` and `detail::delta_unit_t` all call it, and the manipulators do not.

Adding the call is not sufficient, because `strong_t` matches the registered factor *exactly* and the registrations are not spelled consistently:

```cpp
UNIT_ADD(area, square_meters, m2, conversion_factor<std::ratio<1>, dimension::area>)   // structural
UNIT_ADD(area, square_feet,   ft2, squared<feet_>)                                    // manipulator form
```

`squared<meters_>` produces `conversion_factor<ratio<1,1>, dimension_pow<length, ratio<2>>, ratio<0>, ratio<0>>`, which is not the type `square_meters_` is registered against. Applying `strong_t` in the three `_impl`s therefore leaves the asserts above failing (measured).

### What a fix needs

A canonical spelling for a conversion factor, so that the form a manipulator builds and the form a dimension header registers are the same type — then `strong_t` bridges them and the round trip closes by name. That would also improve diagnostics, which currently print a `conversion_factor<...>` where they could print `square_meters`.

There is a hazard to settle first: `strong_t` resolves at instantiation, so a type built inside `squared<>` would depend on which dimension headers are included at that point — `squared<meters_>` would be `square_meters_` in a translation unit that includes `units/area.h` and unnamed in one that does not, which two translation units could disagree about. `common_type` and `delta_unit_t` already carry that exposure; extending it to every squared and rooted type widens it.

### Not a correctness bug

[...]

PR #441, opened 2026-09-11
test: pin the round-trip identity at the call site

nholthaus:
Follow-up to #435, which merged while this was being written.

`detail::rewrap_to_named_t` puts an operation's result back into the named unit, so at the level a caller writes, the type algebra closes for an SI base unit:

```cpp
static_assert(std::is_same_v<units::area::square_meters<double>,
    std::decay_t<decltype(meters<double>(2.0) * meters<double>(3.0))>>);
static_assert(std::is_same_v<meters<double>,
    decltype(units::sqrt(meters<double>(2.0) * meters<double>(2.0)))>);
```

Nothing pinned that, so nothing would notice it breaking.

For a unit whose factor is spelled relative to a parent rather than to the dimension — `feet` is `conversion_factor<ratio<381, 1250>, meters<>>`, and `degrees` is `conversion_factor<ratio<1, 180>, radians<>, ratio<1>>` — the round trip returns by value and by comparison but **not** by name. The test asserts what holds rather than what one would prefer; #440 tracks closing it by name.

405 tests pass on gcc-13.

### DIFF
--- a/test/main.cpp
+++ b/test/main.cpp
@@ -8781,6 +8781,27 @@ TEST(UnitComparison, aWideIntegralOperandOrdersExactly)
 	EXPECT_TRUE(meters<double>(5.0) < kilometers<double>(3.0));
 	EXPECT_TRUE(meters<int>(1000) == kilometers<int>(1));
 }
+// The type algebra closes at the level a caller writes: `detail::rewrap_to_named_t` puts an operation's result back
+// into the named unit, so a product of two lengths IS `square_meters<double>` and the root of it IS `meters<double>`.
+// The round trip is an identity of type as well as value, and needs no annotation to store.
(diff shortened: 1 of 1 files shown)


nholthaus:
Once #442 merges I will rebase this onto main and rewrite its assertions to use `traits::is_same_unit_v`, which is the trait this test was hand-rolling. Verified against #442's branch already: 

- `is_same_unit_v<feet<double>, decltype(sqrt(ft * ft))>` and the `degrees` equivalent both **pass**, where `std::is_same_v` reports them as different types — so every unit gets the same assertion instead of feet and degrees being downgraded to a value-and-comparison check.
- The `std::is_same_v` asserts stay for the SI base units, where the result really is the named type exactly.

Net effect: one uniform contract across metres, feet, kelvin and degrees, and the test states the contract rather than which units happen to resolve to a named type.

PR #442, opened 2026-09-11
feat: ask whether two units are equivalent by the library's own definition

nholthaus:
`std::is_same_v` is the wrong question for a unit, and the tests in #435 and #441 had to work around it by hand — asserting the conversion ratio, the dimension, the pi exponent, the datum and convertibility one at a time, because no single trait said "these are one unit."

A named unit is a class deriving from its `unit<...>`, and a conversion factor has a named spelling and a structural one, so one unit written two ways reports as two types:

```cpp
static_assert(!std::is_same_v<meters<double>, unit<units::length::meters_, double>>);       // two types
static_assert(traits::is_same_unit_v<meters<double>, unit<units::length::meters_, double>>); // one unit
```

That is also what makes the round trip through `squared` look broken when it isn't:

```cpp
static_assert(traits::is_same_unit_v<feet<double>, unit<square_root<squared<units::length::feet_>>, double>>);
static_assert(traits::is_same_unit_v<units::angle::degrees<double>,
    unit<square_root<squared<units::angle::degrees_>>, double>>);
```

### What it adds

- `traits::is_same_conversion_factor<Cf1, Cf2>` / `_v` — the four parts a conversion is built from: dimension, conversion ratio, pi exponent, datum.
- `traits::is_same_unit<U1, U2>` / `_v` — the above plus the numerical scale.
- `units::same_unit<UnitTo, UnitFrom>` — the concept form.

Mirrors the existing `is_same_dimension_conversion_factor` / `is_same_dimension_unit` pair and the `same_dimension` concept beside them, in the same style and the same places.

### What is and is not part of a unit's identity

The **numerical scale is**. `UNIT_ADD_DECIBEL` builds `dBW` from `watts`'s conversion factor, so those two share every part of that factor and are still not the same unit — three of one is not three of the other:

```cpp
static_assert(traits::is_same_conversion_factor_v<units::power::dBW<double>::conversion_factor,
                                                 units::power::watts<double>::conversion_factor>);
static_assert(!traits::is_same_unit_v<units::power::dBW<double>, units::power::watts<double>>);
```

The **representation is not**: `meters<double>` and `meters<int>` are the same unit.

### Verification

Three tests. One unit written two ways answers yes, including the round trip for a metre, a foot, a degree and a kelvin. A difference in any one of the five parts answers no — a different dimension, ratio, pi exponent, datum (celsius against kelvin, which share a dimension *and* a ratio and differ only by 273.15) or scale. And the concept agrees with what conversion actually does.

407 tests pass on gcc-13 and clang-19, warning-free. 77/77 diagnostic cases. Doxygen clean. Reference docs for both traits and the concept, and one changelog bullet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project are documented here. The format follows
 
 ## [Unreleased]
 
+### Added
+- `units::traits::is_same_unit`, `units::traits::is_same_conversion_factor` and the `units::same_unit` concept, which
+  ask whether two types are one unit by the library's definition -- the same conversion factor and numerical scale --
+  where `std::is_same_v` reports a named unit and the `unit<...>` an operation builds for it as different types.
+
 ### Fixed
 - `squared`, `cubed` and `square_root` drop the datum, as their documentation states and as `operator*` already did, so
   the manipulator spelling and the product of two readings are one type and a value carries between the two spellings of
--- a/docs/reference/concepts.md
+++ b/docs/reference/concepts.md
@@ -22,6 +22,7 @@ using namespace units;
 | `DimensionedUnitType<T>` | a unit that **has** a dimension (excludes dimensionless) | `is_unit_v && !is_dimensionless_unit` (`DimensionedUnitType` in `include/units/core.h`) |
 | `DimensionlessUnitType<T>` | a **dimensionless** unit (`dimensionless`, `percent`, angle ratios, …) | `is_unit_v && is_dimensionless_unit` (`DimensionlessUnitType` in `include/units/core.h`) |
 | `same_dimension<UnitTo, UnitFrom>` | two units of the **same dimension** (mutually convertible, e.g. `meters` and `feet`) | `traits::is_same_dimension_unit` (`same_dimension` in `include/units/core.h`) |
+| `same_unit<UnitTo, UnitFrom>` | two types which are the **same unit** — the same conversion factor and numerical scale, so a value converts between them unchanged | `traits::is_same_unit` (`same_unit` in `include/units/core.h`) |
 | `RatioDimensionlessUnitType<U>` | a dimensionless unit whose conversion ratio is **not** 1 — `percent`, `ppm`, `ppb`, … | `traits::is_ratio_dimensionless_cf_v` (`RatioDimensionlessUnitType` in `include/units/core.h`) |
 | `OrdinaryDimensionlessUnitType<U>` | a dimensionless unit whose ratio **is** 1 — plain `dimensionless` (the complement of `RatioDimensionlessUnitType`) | (`OrdinaryDimensionlessUnitType` in `include/units/core.h`) |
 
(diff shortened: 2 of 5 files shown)


chiphogg:
This is a great feature, giving end users a tool to ask what they really _mean_.  I wanted to raise some questions on the name, because now's the best chance to make sure you end up with the name you're most happy with.

Do you want to call this the "same" unit, or an "equivalent" unit?  And if the latter, should we further distinguish between "quantity equivalent" and "point equivalent" units?

These questions are born from my Au experience.  In particular, with Au:

- "meter hertz" and "meters per second" are not the _same_ unit, but they _are_ "quantity equivalent" units.
- Kelvins and degrees Celsius are "quantity equivalent" units, but _not_ "point equivalent" units.

Just offering these ideas in case you find them helpful.  You'll of course know what makes the most sense within the idioms of your library.

Cheers, and keep up the great work!  It's exciting to see nholthaus units roaring back to very active development. 🚀

nholthaus:
@chiphogg thank you — renamed, and you were right for a reason I hadn't measured.

**On "same" versus "equivalent".** Your metre-hertz example decided it. I made it a test:

```cpp
static_assert(traits::is_equivalent_unit_v<meterHertz, mps>);   // true
static_assert(!std::is_same_v<meterHertz, mps>);                // false
```

The trait deliberately ignores how a factor was composed, so "same" overstated it — and `std::is_same_v` already means same, so it was the wrong word to spend. Now `is_equivalent_conversion_factor`, `is_equivalent_unit`, and the `equivalent_unit` concept, with the doxygen naming both boundary cases outright: composition is not part of the comparison, the datum is.

Your question also turned up something. The test suite had been hand-rolling this predicate for **38 assertions** under the name `has_equivalent_conversion_factor` — comparing the dimension and the conversion ratio only. So your vocabulary was already the library's; it just wasn't a public trait. That lambda now calls the trait, and two of the 38 were passing while asserting something untrue: that `common_type_t<kelvin, celsius>` is equivalent to *celsius*. It carries no datum, so it is equivalent to the kelvin — kelvin and celsius share a dimension and a ratio, which is all the old check looked at, so it passed against either. Fixed, with the negative case pinned beside it.

**On quantity- versus point-equivalence,** I'd like to push back gently, because this library resolves it differently than Au. Here the point/amount distinction is carried by the *type* rather than by the predicate — a reading and its difference are different types:

```cpp
static_assert(!traits::is_equivalent_unit_v<celsius<double>, kelvin<double>>);   // a reading is not its amount
using celsiusDifference = decltype(celsius<double>(30.0) - celsius<double>(10.0));
static_assert(traits::is_equivalent_unit_v<celsiusDifference, kelvin<double>>);  // but its difference is a kelvin
```

So one predicate applied to whichever the caller holds gives the right answer, and a second would restate what the types already say. A detail that fell out and pleased me: the difference of two *fahrenheit* readings is equivalent to a **rankine**, not a kelvin — five ninths of the size — so it discriminates on degree size and not only on origin. Both are now tests.

If a call site ever wants "same magnitude, ignore origin" explicitly, it is `is_equivalent_unit_v<differenceOfA, differenceOfB>`, and `detail::delta_unit_t` already names that strip. I'd rather add the second trait when something needs it than ship a taxonomy ahead of its callers.

Much appreciated — this landed at exactly the right time, and the 38-assertion find was worth the whole exchange.

chiphogg:
Awesome --- I'm glad my comments were able to help! 🙂

nholthaus:
Yes, thanks! I agree with your points about naming.

